PR #126, opened 2013-01-16
Handle url query response with text/plain Content Type

tagCincy:
IBM's Smartcloud for Social Business [OAuth 2.0 Implementation](http://www-10.lotus.com/ldd/appdevwiki.nsf/xpDocViewer.xsp?lookupName=IBM+Social+Business+Toolkit+documentation#action=openDocument&res_title=OAuth_2.0_APIs_for_web_server_flow_sbt&content=pdcontent) returns the access token response as a query string but with a content type of 'text/plain'.

A minor change to check a plain text result to see if it is a query:

``` ruby
if @content_type.eql?('text/plain') 
 @content_type = response.body.match(/([a-zA-Z0-9_]+\=[a-zA-Z0-9_]+\&*)+/) ? 'application/x-www-form-urlencoded' : 'text/plain'
else
 @content_type
end
```

This will allow for similar implementations to parse the response correctly.


spec/oauth2/response_spec.rb
```diff
@@ -60,6 +60,25 @@
...
+      expect(subject.parsed['answer']).to eq('42')
+    end
+
+    it "parses text/plain that isn't a url query" do
+      headers = {'Content-Type' => 'text/plain'}
+      body = 'foo equals bar'
+      response = double('response', :headers => headers, :body => body)
+      subject = Response.new(response)
+      expect(subject.content_type).to eq('text/plain')
+    end
```
pboling: There may be merit to the core idea of this PR.  I don't really know if handling plain text response is in the scope of oauth2 though.

Issue #156, opened 2013-10-30
Suggestion: Allow lambdas for `parse` option


niels:
We noticed an inconsistency in how Facebook responds to OAuth requests during callbacks.

When sending an access token, Facebook responds with a Content-Type of `text/plain` while the body is actually `application/x-www-form-urlencoded` encoded. That's why mkdynamic/omniauth-facebook hardcodes the `parse` option to `:query`.

During error conditions however, Facebook responds with `application/json` and the body being a JSON encoded Hash giving more information about the error encountered. Due to the need for hardcoding the parser to query for the normal responses (as Facebook misreports the content type), the error responses are now no longer parsed correctly.

To test this, simply attempt two OAuth requests against facebook. One using correct application credentials (thus receiving the form encoded response) and one using a wrong application secret (thus receiving a JSON encoded error response).

Currently, I don't see a clean way to institute different parsing behaviours for different responses without monkey-patching. One has to either always rely on the reported content type or always use the same parser as defined via the `parse` option. I would thus like to suggest having the latter accept lambdas which would get passed the entire response (incl. headers) and return :query, :json, etc. as appropriate. The behaviour for non-lambda values would remain the same, thus ensuring backward compatibility.

Let me know what you think; I'd be happy to send a pull request.

niels:
Closing: #158 addresses this use-case. See https://github.com/oauth-xx/oauth2/issues/156#issuecomment-29013524 for usage information.

pboling:
Documented in https://github.com/oauth-xx/oauth2/wiki/Examples#omniauth-facebook

PR #158, opened 2013-11-21
Optionally pass raw response to parsers

niels:
This addresses the use case described in #156 and should be fully backwards compatible.

Tested against MRI 1.8.7-p374, 1.9.2-p320, 1.9.3-p448, 2.0.0-p247 and jruby-1.7.4.

### DIFF
--- a/lib/oauth2/response.rb
+++ b/lib/oauth2/response.rb
@@ -67,8 +67,13 @@ module OAuth2
     #   Will attempt to parse application/x-www-form-urlencoded and
     #   application/json Content-Type response bodies
     def parsed
-      return nil unless @@parsers.key?(parser)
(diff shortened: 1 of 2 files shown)


pboling:
@niels Sorry for the delays.  This gem is now being maintained be a *new* team of Awesomenauts.

If you can, we would appreciate another rebase.  We are pushing hard for a final 1.X series release, and then a 2.0 release to modernize the gem (drop Ruby < 2.2).

niels:
Sure thing; very glad to see this project being picked up again! Many thanks to the new maintainers.

Patch rebased and coding style slightly improved.

Tested against MRI 2.5.0, 2.4.2, 2.3.6 & 2.2.7.

PR #159, opened 2013-12-19
added support gzip content encoding

stavskiys:
I use omniauth-facebook gem. For some users facebook return gzipped body. In this case user can't work on site under fb account.


ChaosSteffen:
This issue still exists with ruby 2.2.5. Please fix it!

josephpage:
#256 fixes the same issue

PR #160, opened 2013-12-23
Put required response_type parameter

brauliobo:
The required response_type parameter is missing (ref. http://tools.ietf.org/html/rfc6749#section-3.1.1)


pboling:
This is now handled in the current version of the code as follows:
```
    class AuthCode < Base
      # The required query parameters for the authorize URL
      #
      # @param [Hash] params additional query parameters
      def authorize_params(params = {})
        params.merge('response_type' => 'code', 'client_id' => @client.id)
      end
```
as per: https://tools.ietf.org/html/rfc6749#section-4.1.1

PR #165, opened 2014-02-03
Add handler for application/hal+json

fancyremarker:
This adds a handler for token server responses of content type `application/hal+json`. I followed the pattern from 56f9758d961fc317a05b2f93fd039588967a84b1 for handling multiple synonymous MIME types. Let me know if this is okay.

I can add in additional JSON extension formats, if you think it makes sense. I omitted Collection+JSON and JSON API because they aren't compatible with the OAuth 2.0 spec.


pboling:
Fixed by #364

PR #166, opened 2014-02-08
Update Assertion strategy to latest specs

fancyremarker:
The JWT bearer assertion strategy is no longer described by the OAuth 2.0 core spec, as it was in v10 of the draft. Instead, a separate "JSON Web Token (JWT) Profile" [draft](http://tools.ietf.org/html/draft-ietf-oauth-jwt-bearer-07) covers this strategy.

The two significant changes are that `sub` is now used in place of `prn` to refer to the principal subject of the assertion, and the assertion type is now described by two parameters, `grant_type` and `assertion_type`, instead of just `grant_type`.

I also allowed the algorithm to be overridden via `params[:algorithm]`, in case for example `RS512` should be used.

### DIFF
--- a/lib/oauth2/strategy/assertion.rb
+++ b/lib/oauth2/strategy/assertion.rb
@@ -4,7 +4,7 @@ module OAuth2
   module Strategy
     # The Client Assertion Strategy
     #
-    # @see http://tools.ietf.org/html/draft-ietf-oauth-v2-10#section-4.1.3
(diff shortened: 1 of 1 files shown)


josephpage:
Hey @fancyremarker,
Sorry for the delay...   The gem now has a new team of maintainers. 🎉 
We will be happy to merge your contribution when the conflicts will be solved and if you can re-check this PR against the final version of the RFC : https://tools.ietf.org/html/rfc7523
Also, we will appreciate some specs to test the feature. Don't hesitate to ask for help.

pboling:
Closed by #220

Issue #116, opened 2012-08-17
OAuth2::AccessToken#refresh! returns new object, not itself 


akostrikov:
It is a common practice that when we use bang! methods methods are performed to object itself.

On line https://github.com/intridea/oauth2/blob/master/lib/oauth2/access_token.rb#L99 #refresh! method returns new token, so we can not use it in such way:

```
def try_refresh
  # some work
  token.refresh!
  # more work
  token
end
```

Instead we should use temporary variable, like this:

```
def try_refresh
  # some work
  new_token = token.refresh!
  # more work
  new_token
end
```

pboling:
This was just fixed, and the method has been renamed to `refresh`, with an alias to `refresh!` for backwards compatibility.  https://github.com/oauth-xx/oauth2/pull/363

PR #169, opened 2014-04-10
Rename AccessToken#refresh! to refresh, leaving a compatibility alias.

### LINKED ISSUE #85: access_token.refresh! should update access_token's token
In current implementation, I get a `OAuth2::AccessToken` instance with does not have a `refresh_token`, so I've to manually update either the first one's `access_token` or the second one's `refresh_token`.

Ideally we should be able to do `access_token.refresh!` and not deal with new instances.

Thanks!

knu:
This fixes #85 and #116.

Suffixing the method name with a bang may lead to a misconception that it may update the receiver especially when OAuth2::AccessToken is partially mutable.

I actually mistook it, and the following is a piece of my updated code.

```
  def request(*args)
    refreshed = false
    begin
      response = @token.request(*args)
    rescue OAuth2::Error => e
      raise if refreshed || e.response.parsed['errorCode'] != 401
      @token = @token.refresh!
      refreshed = true
      retry
    end
    # ...
  end
```

The `@token = @token.refresh!` part looks unfamiliar to me as a long time ruby programmer.

### DIFF
--- a/lib/oauth2/access_token.rb
+++ b/lib/oauth2/access_token.rb
@@ -77,7 +77,7 @@ module OAuth2
     #
     # @return [AccessToken] a new AccessToken
     # @note options should be carried over to the new AccessToken
-    def refresh!(params = {})
+    def refresh(params = {})
       fail('A refresh_token is not available') unless refresh_token
       params.merge!(:client_id      => @client.id,
                     :client_secret  => @client.secret,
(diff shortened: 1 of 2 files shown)


pboling:
The source repository for this PR disappeared so I wasn't able to maintain the actual author by rebasing it onto latest master.  I've reimplemented these changes myself in #363

PR #175, opened 2014-05-11
Allow alternate Authorization header field names

shaabans:
Working with SugarCRM's REST API requires an authorization field header of "OAuth-Token" instead of "Authorization".  I needed to be able to do something like this (last line below is the important one, where I just added a new option, header_auth_field):

``` ruby
client = OAuth2::Client.new client_id, client_secret, :token_url => '/rest/v10/oauth2/token', :site => client_url
token_request = client.password.get_token(username, password)
token_string = token_request.token
request = OAuth2::AccessToken.new client, token_string, :header_format => "%s", :header_auth_field => "OAuth-Token"
```


pboling:
@shaabans did SugarCRM fix their header detection?  Is there still an immediate need for this to make SugarCRM work?

PR #176, opened 2014-05-16
Update specs to RSpec 2.14.8 syntax

loganhasson:
This conversion is done by Transpec 1.13.1 with the following command:
    transpec
- 2 conversions
  from: obj.stub(:message)
    to: allow(obj).to receive(:message)


josephpage:
This PR is no longer relevant as faulty lines no longer exist.

PR #178, opened 2014-05-21
Omit extra params when Authorization header is present

jcromartie:
The OAuth 2 spec states that a token request with extra parameters or multiple methods of authorizing the client (e.g. both an "Authorization: Basic" header and client_id and client_secret params) should be treated as an invalid_request error.


pboling:
This was replaced by #280

PR #179, opened 2014-05-21
Adds the :client_auth option

jcromartie:
This is an alternative to the other pull request that's open which removes extra params when the Authorization header is present.

The solution in this branch requires the user to be explicit and pass the option :client_auth => :http_basic when initializing the client in order to attach HTTP Basic auth to the token request and other requests.


josephpage:
Thanks @jcromartie 
Already implemented in 1.3.0.

PR #191, opened 2014-09-07
Token is expired if 'expired_at' time is now.

davestevens:
The `expired?` method on `AccessToken` does not take into account the current time. If `expires_at` is right now `expired?` currently returns `false`. Is there a reason for this?
It means I can not catch a token refresh without using a `rescue` for `OAuth2::Error` after my request then then refreshing the token (which may not always be the cause of the error.

Example code of issue:

``` ruby
require "oauth2"

# Rails app using Doorkeeper with `access_token_expires_in` set to `2.seconds`
client_id = "abc"
client_secret = "def"
endpoint = "https://api.example.com"

client = OAuth2::Client.new(client_id, client_secret, site: endpoint)
token = client.password.get_token("username", "password")
puts token.get("/version").body

# Sleep until the Token is expired
sleep token.expires_at - Time.new.to_i
# Expect this to return true
puts token.expired? # Prints false
token = token.refresh! if token.expired?

# This request fails due to expired Token
puts token.get("/version").body
```

### DIFF
--- a/lib/oauth2/access_token.rb
+++ b/lib/oauth2/access_token.rb
@@ -70,7 +70,7 @@ module OAuth2
     #
     # @return [Boolean]
     def expired?
-      expires? && (expires_at < Time.now.to_i)
+      expires? && (expires_at <= Time.now.to_i)
     end
 
     # Refreshes the current Access Token
--- a/spec/oauth2/access_token_spec.rb
+++ b/spec/oauth2/access_token_spec.rb
@@ -133,6 +133,13 @@ describe AccessToken do
       expect(access).to be_expired
     end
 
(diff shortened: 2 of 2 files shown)


josephpage:
Hey @davestevens,
Sorry for the delay...   The gem now has a new team of maintainers 🎉 

You're 100% right !
I will be happy to merge your brilliant contribution when the conflict will be solved.
Thanks !

davestevens:
@josephpage I've resolved the conflict (as far as I'm aware) through the github UI, although its still being shown as conflicting? I'll have to check later.

josephpage:
It's ok. Seems Github has a issue when conflicts are resolved in web UI.
Thanks !

PR #192, opened 2014-09-12
auth_code: default to header based authorization

maletor:
http://tools.ietf.org/html/rfc6749#section-2.3.1

> Including the client credentials in the request-body using the two
>    parameters is NOT RECOMMENDED and SHOULD be limited to clients unable
>    to directly utilize the HTTP Basic authentication scheme (or other
>    password-based HTTP authentication schemes).  The parameters can only
>    be transmitted in the request-body and MUST NOT be included in the
>    request URI.

### DIFF
--- a/lib/oauth2/strategy/base.rb
+++ b/lib/oauth2/strategy/base.rb
@@ -11,6 +11,14 @@ module OAuth2
       def client_params
         {'client_id' => @client.id, 'client_secret' => @client.secret}
       end
+
+      # Returns the Authorization header value for Basic Authentication
+      #
+      # @param [String] The client ID
+      # @param [String] the client secret
+      def authorization(client_id, client_secret)
+        'Basic ' + Base64.encode64(client_id + ':' + client_secret).gsub("\n", '')
+      end
     end
   end
 end
--- a/lib/oauth2/strategy/base.rb
+++ b/lib/oauth2/strategy/base.rb
@@ -9,7 +9,12 @@ module OAuth2
       #
       # @return [Hash]
       def client_params
-        {'client_id' => @client.id, 'client_secret' => @client.secret}
+        case @client.options[:auth_scheme]
+        when :request_body
+          {'client_id' => @client.id, 'client_secret' => @client.secret}
+        when :basic_auth
+          {:headers => {'Authorization' => authorization(@client.id, @client.secret)}}
+        end
       end
 
       # Returns the Authorization header value for Basic Authentication
--- a/lib/oauth2/strategy/base.rb
+++ b/lib/oauth2/strategy/base.rb
@@ -12,7 +12,7 @@ module OAuth2
         case @client.options[:auth_scheme]
         when :request_body
           {'client_id' => @client.id, 'client_secret' => @client.secret}
-        when :basic_auth
+        else
           {:headers => {'Authorization' => authorization(@client.id, @client.secret)}}
         end
       end
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -1,5 +1,6 @@
 require 'faraday'
 require 'logger'
+require 'byebug'
 
 module OAuth2
   # The OAuth2::Client class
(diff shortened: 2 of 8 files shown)


aviscasillas:
Hi!
Thinking on a solution to send client credentials using basic authentication, I've found this PR. I see that these changes are basically what I'm looking for, and I'm wondering if it will go forward (I see the last comment is old, but the PR is still open :confused:).
Thx!

recmend:
Any update on this? This impacts Yahoo OAuth 2.0 integration. Yahoo refuses to permit client credentials in the request body.

stevenringo:
Please can this be merged?

sferik:
> I leave it up to you how you want to handle it, but basically our options are to:
> 1. Include a deprecation message in the get_token method here and support it until some time in the future.
> 2. Remove support (as it currently is) and major or minor version bump depending on how you feel about supporting an undocumented feature.

My preference would be for option 1 to give users a easier migration path.

@maletor Would you be willing to update your patch to deprecate but continue to support the existing behavior?

That is my only objection to merging and releasing this patch.

maletor:
Sure, I can update the PR to deprecate it. I'll look at it sometime today.

sferik:
@maletor Awesome! Thanks!

stevenringo:
These changes also need to filter through to `AccessToken#refresh!`, which still uses the `:request_body` method. This looks like it could be a messy (re)factoring as `Strategy::Base#client_params` has the logic that's required in `AccessToken`.

stevenringo:
I would like to make changes to this PR, to add more address the issue I raised in the previous comment. Do I create a new PR that incorporates this PR's changes, plus mine, or is there a way I can update this PR?

maletor:
Feel free to fork my branch of cherry pick my commits. I agree, if we are
going to deprecate, it's going to require significantly more changes.

etisab:
Hello, I was wondering what the status of these changes are and if they will they be merged in the near future. We are also doing Yahoo! OAuth 2.0 integration which requires authorization credentials in the request header. Thanks

maletor:
Any updates on this?

josephpage:
[Xee API](https://github.com/xee-lab/documentation) also requires authorization credentials in the request header for both get and renew access token.
Need this to be merged ! 🙏  @sferik

bjeanes:
I also have an OAuth provider which rejects authentication other than by `Basic` for exchanging tokens. Does this need help getting it over the line or is there a bigger picture change in mind instead?

ozgg:
So... is there any chance this will be merged at all?

sferik:
This pull request no longer merges cleanly.

lib/oauth2/strategy/base.rb
```diff
@@ -11,6 +11,14 @@ def initialize(client)
       def client_params
         {'client_id' => @client.id, 'client_secret' => @client.secret}
       end
+
+      # Returns the Authorization header value for Basic Authentication
+      #
+      # @param [String] The client ID
+      # @param [String] the client secret
+      def authorization(client_id, client_secret)
+        'Basic ' + Base64.encode64(client_id + ':' + client_secret).gsub("\n", '')
```
bjeanes: @divoxx I don't think that is correct in the case of the `Authorization` header, as it is Base64 encoded. I wasn't sure so I tested to see what `curl` does here. Furthermore, the RFC does not specify any encoding other than Base64: https://tools.ietf.org/html/rfc2617#section-2.

bjeanes: In my experiments with authenticating with a few OAuth providers, it seems that at least some reject form-encoded credentials. I have one with a `!` in the password, which would get encoded as `%21`, but this is rejected.

ozgg:
Should @maletor or anyone else with write access do this? Because there are providers that fail to work with current version of the gem without these changes.

bjeanes:
@sferik @ozgg I am currently preparing a new PR. ETA ~1 day

bjeanes:
(I also have another PR in prep to bring the lib into spec around [redirect_uri for access tokens](https://tools.ietf.org/html/rfc6749#section-4.1))

Issue #197, opened 2014-09-16
Error in generated body with assertion strategy


fabn:
I cannot replicate what [google-api-ruby-client](https://github.com/google/google-api-ruby-client) is doing (via [signet oauth client](https://github.com/google/signet)).

I'm using a [service account](https://developers.google.com/accounts/docs/OAuth2ServiceAccount) to access google analytics api and this code, taken from a google example, is working

``` ruby
Signet::OAuth2::Client.new(
                :token_credential_uri => 'https://accounts.google.com/o/oauth2/token',
                :audience => 'https://accounts.google.com/o/oauth2/token',
                :scope => 'https://www.googleapis.com/auth/analytics.readonly',
                :issuer => SERVICE_ACCOUNT_EMAIL_ADDRESS,
                :signing_key => Google::APIClient::KeyUtils.load_from_pkcs12(PATH_TO_KEY_FILE, 'notasecret'),
                :person => USER_ACCOUNT,
            )
client.fetch_access_token!
```

Since I'm already using oauth2 gem in my project I'd like to avoid inclusion of another oauth library, so I tried to translate the above code to work with oauth2, here's my attempt:

``` ruby
# Oauth client, client id and secret not needed
client = OAuth2::Client.new('', '', token_url: 'https://accounts.google.com/o/oauth2/token')
# Build params for JWT
params = {
    private_key: Google::APIClient::KeyUtils.load_from_pkcs12(PATH_TO_KEY_FILE, 'notasecret'),
    aud: 'https://accounts.google.com/o/oauth2/token',
    scope: 'https://www.googleapis.com/auth/analytics.readonly',
    iss: SERVICE_ACCOUNT_EMAIL_ADDRESS, # Service account email
    prn: email, # Acts as this user
    exp: Time.now.utc.to_i + 60,
}
# Return the access token
client.assertion.get_token(params) # => this code raise an error
#       OAuth2::Error:
#      invalid_grant:
#     {
#         "error" : "invalid_grant"
#      }
```

If I configure oauth client to dump all requests (using excon) I get this request:

```
excon.request
  # notice this is an hash
  :body              => {
    :assertion      => "...JWT Encrypted assertion..."
    :assertion_type => "urn:ietf:params:oauth:grant-type:jwt-bearer"
    "client_id"     => ""
    "client_secret" => ""
    :grant_type     => "assertion"
    :scope          => "https://www.googleapis.com/auth/analytics.readonly"
  }
  }
```

[...]

PR #198, opened 2014-09-17
Assertion Strategy should allow any param in claim set

fabn:
Related to #197, allow any param for `build_assertion` method.

Let me know if this need something else to get merged.


pboling:
This will be fixed by #220 .

Thanks for your work on this @fabn

PR #201, opened 2014-09-30
Remove circular reference for OAuth2::Error

brainopia:
OAuth2::Error contains response and adds a link to itself from response. This is a circular reference and it can lead to various problems. E.g., attempt to dump exception to json (that most exception handlers are doing) will lead to `ActiveSupport::JSON::Encoding::CircularReferenceError: object references itself`.

This is why I propose to remove this inter-dependency.

[Review on Reviewable](https://reviewable.io/reviews/intridea/oauth2/201)

### DIFF
--- a/lib/oauth2/error.rb
+++ b/lib/oauth2/error.rb
@@ -5,7 +5,6 @@ module OAuth2
     # standard error values include:
     # :invalid_request, :invalid_client, :invalid_token, :invalid_grant, :unsupported_grant_type, :invalid_scope
     def initialize(response)
-      response.error = self
       @response = response
 
       message = []


pboling:
I will work on a spec for this issue.

jhmoore:
So, I think this is a pretty reasonable thing to not have a circular reference between two objects.  However:
 1. As you say, tests -- it's a little alarming that no tests failed removing that line (although haha as of https://github.com/oauth-xx/oauth2/pull/380 there are now actual tests for `OAuth2::Error` :suspect: )
 2. If we're going to be removing this behavior, we should probably remove the `attr_accessor :error` on `OAuth2::Response` because now it will never be set.

We could also solve this problem the other way around -- i.e. removing the `response` from `OAuth2::Error` -- but either way it'd be a breaking change.  Which do we think is a more common use case?  Calling `response.error` or `error.response`?

jhmoore:
Hah.  Looking at this more, this PR might also not even fix the issue.  In `OAuth2::Client`:
https://github.com/oauth-xx/oauth2/blob/master/lib/oauth2/client.rb#L115-L119
```ruby
      when 400..599
        error = Error.new(response)
        raise(error) if opts.fetch(:raise_errors, options[:raise_errors])
        response.error = error
        response
```

So even if that line in `OAuth2::Error` was removed, we'll still be setting the instance of the error to the attr on the response manually in the client.  

Haha this does also raise the point that that line of code in `Client` is pretty redundant -- it's initializing `Error`, which right now is automatically doing the attr setting anyway.

pboling:
Breaking changes are allowed in the next release, as it will be a major release.
We should come up with a *better* API than we have now for this situation.  @jhmoore makes a lot of good points.

jhmoore:
@mathieujobin thanks for reminding me about this.  I don't believe this PR will fix the underlying issue described, see above ⬆️ there's another place that `response.error` gets set.  I can take a crack at removing all instances of setting `Error` on `Response`.

jhmoore:
(Additionally, this _will_ make tests fail on master, because I backfilled specs for existing error class behaviors in https://github.com/oauth-xx/oauth2/pull/380)

PR #202, opened 2014-10-08
Fixing camel case content types that providers sometimes reply with

eladmeidar:
Add methods that standardise the response JSON keys into snake cased keys. we tried working with an oauth provider that used the java representation form "accessToken" instead of "access_token".

other thing is that we downcased the content_type header value to match parser matching.


pboling:
Closing in favor of https://github.com/oauth-xx/oauth2/pull/507

PR #204, opened 2014-10-22
fix for Faraday::RackBuilder::StackLocked when logging debug messages

james-lawrence:
faraday raises "Faraday::RackBuilder::StackLocked (can't modify middleware stack after making a request)"
if you try to add the logger in the request method.


pboling:
@james-lawrence apologies for not seeing this earlier (or maybe I did see it and I used it as inspiration, I don't recall!).  I included a change that effectively does this same thing in this major overhaul PR: https://github.com/oauth-xx/oauth2/pull/346

Thanks for your contribution.

PR #205, opened 2014-10-28
Adds support for handling respones from Collection+JSON APIs

ryanwi:
For APIs using the [Collection+JSON Hypermedia document format](http://amundsen.com/media-types/collection/), responses are not being parsed. Collection+JSON is still JSON, so this PR connects responses with the Collection+JSON content type to the JSON parser.


pboling:
New PR with your changes rebased on latest master is #365

PR #212, opened 2015-03-19
Request client.get_token does not use headers in methods other then POST

MikhailZhukov:
Request does not use headers in methods other then POST. Basic authorization does not work with GET method.

### DIFF
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -127,14 +127,14 @@ module OAuth2
     # @return [AccessToken] the initalized AccessToken
     def get_token(params, access_token_opts = {}, access_token_class = AccessToken)
       opts = {:raise_errors => options[:raise_errors], :parse => params.delete(:parse)}
+      headers = params.delete(:headers) || {}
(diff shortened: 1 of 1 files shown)


pboling:
This has been fixed in the current version of `get_token`:
```
    def get_token(params, access_token_opts = {}, access_token_class = AccessToken) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
      params = authenticator.apply(params)
      opts = {:raise_errors => options[:raise_errors], :parse => params.delete(:parse)}
      headers = params.delete(:headers) || {}
      if options[:token_method] == :post
        opts[:body] = params
        opts[:headers] = {'Content-Type' => 'application/x-www-form-urlencoded'}
      else
        opts[:params] = params
        opts[:headers] = {}
      end
      if options[:raise_errors] && !(response.parsed.is_a?(Hash) && response.parsed['access_token'])
        error = Error.new(response)
        raise(error)
      end
      build_access_token(response, access_token_opts, access_token_class)
    end
```

PR #219, opened 2015-08-21
Loosen rack requirement to allow testing with rack master branch

twalpole:
This loosens the rack version requirement to allow testing with the rack master branch as currently being required by rails 5. It also adds to travis as an allowable failure, and fixes/ignores some rubocop warnings

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 *~
 .bundle
 .rvmrc
-Gemfile.lock
+Gemfile*.lock
 coverage/*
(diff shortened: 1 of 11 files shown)


zernie:
+

MartinodF:
+1, the current strict requirement on rack ~> 1.2 means that not only oauth2 but also omniauth are incompatible with rails 5.

tpmullan:
Why hasn't this been merged?

Issue #203, opened 2014-10-20
JWTs using deprecated 'prn' value


sqrrrl:
Since the -04 revision for the JWT profile, 'prn' has been renamed to 'sub'. Code is still using the old name.

http://tools.ietf.org/html/draft-ietf-oauth-jwt-bearer-04

xtagon:
PR would be welcome

PR #220, opened 2015-08-28
Update Assertion strategy claims and request parameters

jhmoore:
This started long ago as a fix to get this gem working with the Google two-legged OAuth2 service account use case, but evolved into a general fix for several issues.  See the message history here for the complete context, but the short version is that this PR:

 - Updates the `assertion_type` request parameter to `grant_type`.  This was changed in [OAuth2 draft v11](https://tools.ietf.org/id/draft-ietf-oauth-v2-11.html#anchor34) in 2010(!).  See [the final RFC](https://tools.ietf.org/html/rfc6749#section-11.2.2).
 - Allows any string to be set as a JWT claim key, rather than restricting to a specific set of four.  Relevant RFCs are [7519](https://tools.ietf.org/html/rfc7519#section-4.1) and [7523](https://tools.ietf.org/html/rfc7523#section-3).  There are some standard/expected ones (`iss`, `sub`, `aud`, `exp`, etc) but any endpoint can specify a requirement for the presence (or absence!) of any custom key they want.
 - Allows the use of any encoding method supported by the [jwt gem](https://github.com/jwt/ruby-jwt), rather than just HS256 or RS256.  See their documentation for the list.
 - Adds a bunch of specs!

The downside:
 - There are **breaking changes to `Assertion#get_token`**.  Previously, the arguments looked like:
```ruby
get_token(params, opts = {})
```
Now, the arguments are:
```ruby
get_token(claims, encoding_opts, request_opts = {}, response_opts = {})
```
One of the reasons the Google use case did not work before was that Google [requires](https://developers.google.com/identity/protocols/OAuth2ServiceAccount) a custom JWT claim key named `scope`.  However, `params[:scope]` was a special key that was getting put into the request body itself -- having _one_ `params` argument meant that this method could not differentiate between arguments intended to be put into the JWT, and arguments intended to be used elsewhere.  In addition, like in the example of `scope`, there could be collisions between the two concepts.

Showing what an identical call might look like, translated from old to new:

**OLD:**
```ruby
params = {
  :iss => 'oauth@example.com',
  :scope => 'foo',
  :aud => 'https://example.com/oauth2/token',
  :exp => 1.hour.from_now.to_i,
  :prn => 'swayze@roadhouse.example.com',
  :hmac_secret => 'secret!'
}

get_token(params)
```

**NEW:**
```ruby
claims = {
  :iss => 'oauth@example.com',
  :aud => 'https://example.com/oauth2/token',
  :exp => 1.hour.from_now.to_i,
  :prn => 'swayze@roadhouse.example.com',
}

get_token(claims, {:algorithm => 'HS256', :key => 'secret!'}, {:scope => 'foo'})
```
Note that there is now the slight additional overhead of needing to know _which_ encoding method you want to use.  If you were passing in `params[:hmac_secret]` before, this file was inferring `'HS256'` -- if you were passing in `params[:private_key]`, it was inferring `'RS256'`.  See [the list in RFC 7518](https://tools.ietf.org/html/rfc7518#section-3.1) of supported algorithms -- including some that may be more strongly recommended in the future.

It's also worth noting that this PR expressly _does not_ try to validate the presence of claim keys that are listed as required in the JWT for OAuth2 [RFC](https://tools.ietf.org/html/rfc7523#section-3).  There are definitely endpoints that do not conform to that RFC (e.g. I think Google does not, in some cases), and folks should still be able to use this gem for those services.  If there are endpoints that are strict about requiring the RFC-mandated keys, then they will enforce that behavior on their end.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ log/*
 measurement/*
 pkg/*
 rdoc/*
+.idea
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -44,6 +44,9 @@ Lint/UnusedBlockArgument:
     - 'vendor/**/*'
     - '**/.irbrc'
 
+RSpec/NestedGroups:
+  Enabled: false
+
 Style/ClassVars:
   Enabled: false
 
--- a/lib/oauth2/strategy/assertion.rb
+++ b/lib/oauth2/strategy/assertion.rb
@@ -49,10 +49,12 @@ module OAuth2
 
       def build_request(params)
         assertion = build_assertion(params)
-        request_hash = {}
-        request_hash[:grant_type] = params[:grant_type] || 'urn:ietf:params:oauth:grant-type:jwt-bearer'
-        request_hash[:assertion] = assertion
-        request_hash.merge!(client_params)
+        grant_type = params[:grant_type] || 'urn:ietf:params:oauth:grant-type:jwt-bearer'
+        
+        {
+          :grant_type => grant_type,
+          :assertion => assertion
+        }
       end
 
       def build_assertion(params)
--- a/lib/oauth2/strategy/assertion.rb
+++ b/lib/oauth2/strategy/assertion.rb
@@ -51,11 +51,8 @@ module OAuth2
       end
 
       def build_request(params)
-        extract_legacy_params!(params)
-        # ^ This can go away whenever we decide to no longer preserve previous gem behavior of restricting keys
-
-        url_params = params[:url_params]
-        claims = params[:claims]
+        url_params = params[:url_params] || {}
+        claims = params[:claims] || {}
         encoding_options = choose_algorithm!(params)
 
         {
@@ -83,36 +80,6 @@ module OAuth2
 
       def build_assertion(claims, encoding_options)
         JWT.encode(claims, encoding_options[:key], encoding_options[:algorithm])
-      end
-
-      def extract_legacy_params!(params)
-        # In previous versions of this gem, `:scope` defaulted to being in the POST request body outside the claimset.
-        # Additionally, :iss, :aud, :prn, and :exp were the only four keys and always present in the claimset
-        #
-        # This method is intended to preserve that behavior, and can be removed when no longer needed
-        #
-        # If you are seeing a deprecation warning in your app: the `params` argument to Assertion is now split into:
-        # params = { :url_params => {}, :claims => {} }
-
-        params[:url_params] ||= {}
-        params[:claims] ||= {}
-        legacy_url_params = {}
-        legacy_claims = {}
-
-        if params.has_key?(:scope)
-          puts "DEPRECATION WARNING: params[:scope] is defaulted to a header value -- this behavior may go away!"
-          legacy_url_params[:scope] = params.delete(:scope)
-        end
-
-        [:iss, :aud, :prn, :exp].each do |legacy_key|
-          if params.has_key?(legacy_key)
-            puts "DEPRECATION WARNING: params[#{legacy_key}] is a legacy default claim -- this behavior may go away!"
-            legacy_claims[legacy_key] = params.delete(legacy_key)
-          end
-        end
-
-        params[:url_params].merge!(legacy_url_params)
-        params[:claims].merge!(legacy_claims)
       end
     end
   end
--- a/lib/oauth2/strategy/assertion.rb
+++ b/lib/oauth2/strategy/assertion.rb
@@ -36,9 +36,12 @@ module OAuth2
       #   params :hmac_secret, secret string.
       #   params :private_key, private key string.
       #
+      # for possible claim keys, see https://tools.ietf.org/html/rfc7519#section-4.1
+      #
       #   params :iss, issuer
       #   params :aud, audience, optional
       #   params :prn, principal, current user
+      #     ^ DEPRECATED: prn is now 'sub' https://tools.ietf.org/html/draft-ietf-oauth-json-web-token-06#appendix-F
       #   params :exp, expired at, in seconds, like Time.now.utc.to_i + 3600
       #
       # @param [Hash] opts options
@@ -48,33 +51,68 @@ module OAuth2
       end
 
       def build_request(params)
-        assertion = build_assertion(params)
-        grant_type = params[:grant_type] || 'urn:ietf:params:oauth:grant-type:jwt-bearer'
-        
+        extract_legacy_params!(params)
+        # ^ This can go away whenever we decide to no longer preserve previous gem behavior of restricting keys
+
+        url_params = params[:url_params]
+        claims = params[:claims]
+        encoding_options = choose_algorithm!(params)
+
         {
-          :grant_type => grant_type,
-          :assertion => assertion
-        }
+          :grant_type => 'urn:ietf:params:oauth:grant-type:jwt-bearer',
+          :assertion => build_assertion(claims, encoding_options),
+        }.merge(url_params)
       end
 
-      def build_assertion(params)
-        claims = {
-          :iss => params[:iss],
-          :aud => params[:aud],
-          :exp => params[:exp],
-          :scope => params[:scope],
-          :iat => params[:iat]
-        }
+    private
 
-        claims[:prn] = params[:prn] if params.has_key?(:prn)
+      def choose_algorithm!(params)
+        # Right now there is only the choice of HS256 or RS256, but this could be expanded
 
-        if params[:hmac_secret]
-          JWT.encode(claims, params[:hmac_secret], 'HS256')
-        elsif params[:private_key]
-          JWT.encode(claims, params[:private_key], 'RS256')
+        private_key = params.delete(:private_key)
+        hmac_secret = params.delete(:hmac_secret)
+
+        if hmac_secret
+          {:algorithm => 'HS256', :key => hmac_secret}
+        elsif private_key
+          {:algorithm => 'RS256', :key => private_key}
         else
-          raise ArgumentError.new 'Either hmac_secret or private_key is required for JWT encoding!'
+          raise ArgumentError.new(:message => 'Either hmac_secret or private_key is required for JWT encoding!')
         end
+      end
+
+      def build_assertion(claims, encoding_options)
+        JWT.encode(claims, encoding_options[:key], encoding_options[:algorithm])
+      end
+
+      def extract_legacy_params!(params)
+        # In previous versions of this gem, `:scope` defaulted to being in the POST request body outside the claimset.
+        # Additionally, :iss, :aud, :prn, and :exp were the only four keys and always present in the claimset
+        #
+        # This method is intended to preserve that behavior, and can be removed when no longer needed
+        #
+        # If you are seeing a deprecation warning in your app: the `params` argument to Assertion is now split into:
+        # params = { :url_params => {}, :claims => {} }
+
+        params[:url_params] ||= {}
+        params[:claims] ||= {}
+        legacy_url_params = {}
+        legacy_claims = {}
+
+        if params.has_key?(:scope)
+          puts "DEPRECATION WARNING: params[:scope] is defaulted to a header value -- this behavior may go away!"
+          legacy_url_params[:scope] = params.delete(:scope)
+        end
+
+        [:iss, :aud, :prn, :exp].each do |legacy_key|
+          if params.has_key?(legacy_key)
+            puts "DEPRECATION WARNING: params[#{legacy_key}] is a legacy default claim -- this behavior may go away!"
+            legacy_claims[legacy_key] = params.delete(legacy_key)
+          end
+        end
+
+        params[:url_params].merge!(legacy_url_params)
+        params[:claims].merge!(legacy_claims)
       end
     end
   end
--- a/lib/oauth2/strategy/assertion.rb
+++ b/lib/oauth2/strategy/assertion.rb
@@ -15,10 +15,15 @@ module OAuth2
     #     :iss => "http://localhost:3001",
     #     :aud => "http://localhost:8080/oauth2/token"
     #     :sub => "me@example.com",
-    #     :exp => Time.now.utc.to_i + 3600
+    #     :exp => Time.now.utc.to_i + 3600,
     #   }
     #
-    #   access = client.assertion.get_token(claimset, :algorithm => 'HS256', :key => 'secret_key')
+    #   encoding = {
+    #     :algorithm => 'HS256',
+    #     :key => 'secret_key',
+    #   }
+    #
+    #   access = client.assertion.get_token(claimset, encoding)
     #   access.token                 # actual access_token string
     #   access.get("/api/stuff")     # making api calls with access token in header
     #
--- a/lib/oauth2/strategy/assertion.rb
+++ b/lib/oauth2/strategy/assertion.rb
@@ -18,7 +18,7 @@ module OAuth2
     #     :exp => Time.now.utc.to_i + 3600
     #   }
     #
-    #   access = client.assertion.get_token(claimset, 'HS256', 'secret_key')
+    #   access = client.assertion.get_token(claimset, :algorithm => 'HS256', :key => 'secret_key')
     #   access.token                 # actual access_token string
     #   access.get("/api/stuff")     # making api calls with access token in header
     #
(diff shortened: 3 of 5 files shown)


josephpage:
Hey @jhmoore,
Sorry for the delay... Now the gem has a new team of maintainers.
I would like to include your PR in the next release (1.5.0). Can you please re-check your code against the RFC and the Google use case then resolve the conflits ?
Please let me know if you have time for this.

jhmoore:
Wow!  @josephpage sorry, this github notification must've gotten completely buried.  I'd be happy to take a look.

pboling:
I would be super interested in adding a code example for this specific use case to our wiki as well.  This is a common need that Google has failed to adequately address in their own help tools and documentation, IMO.  I think there are a fair number of people wondering how to accomplish this with a service account (me being one of them).

jhmoore:
@pboling good to know.  FWIW we've been using this gem with the added contents of this PR to support both 3-legged and 2-legged OAuth with Google for a while now (haha about 2.5 years I guess!) and it's been rock solid.  I have some time booked on my calendar to give this PR some attention -- I'll try to include some documentation and whatnot as well.

jhmoore:
Alrighty!  I just spent a few hours reading RFCs and all kinds of fun things.  I'm currently working on making sure the branch still works with Google, but I'm slowly giving the entire `Assertion` class an overhaul.

Learnings include:
 - Apparently when we implemented Google OAuth2 a few years ago, they were still using the `prn` claim key to indicate 'principal' -- i.e. "the person on behalf of whom you are requesting access via a service account".  This concept, at some point, [was apparently renamed](https://github.com/google/google-api-nodejs-client/issues/148#issuecomment-38085143) to `sub`, for 'subject'.  (I double-swear though that the Google service account docs at the time had `prn` in them)  The [spec](https://tools.ietf.org/html/rfc7519#section-4.1) confirms this:
```
4.1.2.  "sub" (Subject) Claim

   The "sub" (subject) claim identifies the principal that is the
   subject of the JWT.  The claims in a JWT are normally statements
   about the subject.  The subject value MUST either be scoped to be
   locally unique in the context of the issuer or be globally unique.
   The processing of this claim is generally application specific.  The
   "sub" value is a case-sensitive string containing a StringOrURI
   value.  Use of this claim is OPTIONAL.
```

 - One of the reasons we ran into a fundamental conflict in this file vs what will work with Google is the fact that some of the claim keys that needed to be in the JWT were mysteriously in the _header_.  (e.g. [scope](https://github.com/oauth-xx/oauth2/pull/220/files#diff-1d89449286c7ef4c9fc261237c559369L55))  I remember finding this very confusing.  Google blows up if you send your requested scopes in the header, but it turns out that totally is a valid use case [per the spec](https://tools.ietf.org/html/rfc7519#section-5.3): 
```
   In some applications using encrypted JWTs, it is useful to have an
   unencrypted representation of some claims.  This might be used, for
   instance, in application processing rules to determine whether and
   how to process the JWT before it is decrypted.

   This specification allows claims present in the JWT Claims Set to be
   replicated as Header Parameters in a JWT that is a JWE, as needed by
   the application.
```

 - In order for Google to work, there were some keys that are currently just straight up not present in `Assertion#build_assertion`.  The only claim keys that can be encoded right now are:
```ruby
{
  :iss => params[:iss],
  :aud => params[:aud],
  :prn => params[:prn],
  :exp => params[:exp]
}
```
[Google specifically asks for](https://developers.google.com/identity/protocols/OAuth2ServiceAccount) `iat` ("issued at"), and `scope` (which, mentioned previously, was in the headers instead).  Their example:
```
{
  "iss":"761326798069-r5mljlln1rd4lrbhg75efgigp36m78j5@developer.gserviceaccount.com",
  "scope":"https://www.googleapis.com/auth/devstorage.readonly",
  "aud":"https://www.googleapis.com/oauth2/v4/token",
  "exp":1328554385,
  "iat":1328550785
}
```
Curious what the 'allowed' claim keys were, I went digging around and came up with the [official list](https://www.iana.org/assignments/jwt/jwt.xhtml).  Given that there are a _lot_ in there (and presumably allowed to be even more), the `build_assertion` strategy should probably just pass through whatever the heck arbitrary claim keys anybody would like to assert.

In addition, another of the difficulties we ran into in this file was the fact that `prn` was _always_ included in the JWT.  That's one of the differentiating factors with Google's OAuth:
screen shot 2018-02-23 at 5 44 21 pm

So, always including `sub` (née `prn`) in a claimset will work as long as you're always trying to do 2-legged service account stuff -- but the minute you try to do a thing for a regular old 3-legged `plebe_dude@gmail.com` account, Google blows up because of course that user doesn't have the ability to issue a domain-wide enterprise approval for scopes or do any kind of service account request-on-behalf-of nonsense.  This is why the current state of this PR only includes `prn` in the claimset if the params contain the key `prn`, instead of leaving it around as a nil-value key.

---
Conclusions:
1. This PR as it stands currently will _break_ anybody using the existing (valid!) behavior of including scopes in the header.  I am resolving that currently.
2. If some claim keys need to not be there always, and literally maybe any string is allowed to be a claim key, `build_assertion` should probably just assemble a hash from the params and use whatever keys somebody wants.  I am also resolving this currently!

Stay tuned for an update on this PR; I've just finished resolving merge conflicts and all that.

jhmoore:
Another mystery solved!  

Right now on master, this gem sends:
```ruby
:assertion_type => 'urn:ietf:params:oauth:grant-type:jwt-bearer'
```
In order to get Google to work, we had to change it to:
```ruby
:grant_type => 'urn:ietf:params:oauth:grant-type:jwt-bearer'
```
Exact same value, but with a different key name.  Turns out: when this file was written, [assertion_type was accurate as of OAuth2 draft v10](https://tools.ietf.org/html/draft-ietf-oauth-v2-10#section-4.1.3).  However . . . it was [changed to grant_type in draft v11](https://tools.ietf.org/id/draft-ietf-oauth-v2-12.html#anchor43).  

From the [final RFC](https://tools.ietf.org/html/rfc6749#section-11.2.2) (which indeed makes no mention of `assertion_type`):

```
   o  Parameter name: grant_type
   o  Parameter usage location: token request
   o  Change controller: IETF
   o  Specification document(s): RFC 6749
```

This _will_ be a breaking change, but I believe that `assertion_type` will be simply incompatible going forward.

pboling:
I think that a breaking change for an implementation supporting a "draft" standard is acceptable in a minor patch, because you can also see it as a change to maintain compatibility with the RFC as it became finalized (i.e. not doing the change could be considered a bug?)

What do others think?  I think it would be OK to include this in the planned 1.5.0 release.

galois17:
@pboling I agree that it would be considered a bug at that point... Sounds good to me.

jhmoore:
@pboling / @galois17 thanks!  Also: I've immediately run into a larger existential question here, with regards to assertion.rb.  

If, per the JWT spec, any arbitrary key is totally allowed to be a claim key, and if _also_ per the spec any arbitrary key is totally allowed to be in the header (a la `scope`) . . . it would seem that the current way of initializing the `Assertion` strategy object is somewhat incompatible with the RFC.

Right now, you call `client.assertion.get_token(params)` -- where `params` is a hash passed to all subsequent methods.  How can we tell which keys of that hash are intended for the header, and which keys are intended for the JWT?

For example: in master right now, `params[:scope]` will be stripped out and put into the header.  However, in some situations (i.e. Google), `params[:scope]` will need to go into the JWT claimset.  In yet other situations, you could have both.

In addition, there are some other certain 'special' keys of the `params` hash (e.g. `params[:hmac_secret]` that are used for other purposes entirely.  Haha if there ever ended up being a legitimate claim key that somebody needed to send named `hmac_secret`, I daresay it would be a problem.

The easy button is for `get_token` to take multiple hash arguments (i.e. a `headers` hash that is totally separate from the `claims` hash) . . . but this would obviously have the side effect of a breaking change for anybody currently using this.  In addition, there is currently a secondary `opts = {}` arguments already in existence, and it usually ends in maximum weirdness to try to have multiple empty-hash-defaulting arguments next to eachother, let alone introduce new ones.

To that end, to preserve backwards-compatibility, I propose breaking the `params` argument into two top-level keys:
```ruby
params = {
  :headers => {},
  :claims => {}
}
```
Backwards compatibility can be preserved by intercepting other top-level keys of `params` and interpreting them the way current `master` does.

This does have the downside of affecting the other methods (`build_request` and `build_assertion`) in this file, but:
 - Those methods do not seem to be called anywhere else in this gem
 - Neither of those methods had tests until we [added them in this PR](https://github.com/oauth-xx/oauth2/pull/220/files#diff-92d92dd92d84dbf8ed1dd8be7a4f5b19R40).

To that end, I am going to modify their arguments a little and also make them private.  This could be a breaking change to anybody calling those methods directly from their application, but those methods seem inherently 'private' in their intent, given what this object is trying to accomplish.

pboling:
I like your ideas.  One idea would be to parse the params hash split to headers and claims as done now, by default, and to allow specification of an array of keys in `opts[:headers]` for which of the keys go to `:headers` and `opts[:claims]` for which of the keys go to `:claims`.  That has the advantage of easily allowing a key to be used in both headers and claims without the user needing to stick it in both places.  Maybe that's not really an advantage though?  🗡

jhmoore:
Thanks!  Stylistically, totally agreed -- however, I tried playing around with putting stuff like that in the `opts` arg . . . the issue I ran into is that right now, `opts` does affect anything having to do with building the request/assertion stuff.  It gets passed directly into `OAuth2::Client#get_token` as the arg "`access_token_opts`".

If folks are currently using that `opts` hash expecting it to be all stuff pertaining to wrangling access tokens, I'm not sure A.) in what ways people might be using it right now, and B.) what things we might break by having certain special keys that get ripped out of it.

I'm going to pursue this nested key structure thing and see how it ends up looking . . . haha if it ends up being completely ridiculous I'll go the other route.

pboling:
Ah, yes.  The ripping out keys unexpectedly is a real concern.  Thanks for re-mentioning that, the import hadn't fully clicked the first time!

jhmoore:
Ok, I am back on this today.  More learnings:

The current behavior of the `params[:scope]` argument is _not_ what I thought it was doing.  It is not getting put into the JWT headers.

[Currently](https://github.com/oauth-xx/oauth2/blob/master/lib/oauth2/strategy/assertion.rb#L56), `params[:scope]` is getting put _into the raw request params_.  E.g. right now, let's say you passed in a params hash of:
```ruby
{ :iss => 'aaa', :aud => 'bbb', :scope => 'ccc' }
```

`:iss` and `:aud` would get packaged into the JWT claimset, and then the client would send a POST request to the token endpoint with a **body** of: `assertion=blahblahblah&grant_type=urn:ietf:params:oauth:grant-type:jwt-bearer&scope=ccc`

I had no idea why this would be a top-level request parameter, and like even if the key isn't passed in, I'm pretty sure right now in master an empty `scope=` is tacked on to every single token request made by the gem.  And there were no tests for that behavior.  I was about to just remove it, but . . . I went digging and found (yet another!!!) RFC.  In particular, [RFC 7521](https://tools.ietf.org/html/rfc7521#section-4.1) describes:
```
scope
      OPTIONAL.  The requested scope as described in Section 3.3 of
      OAuth 2.0 [RFC6749].  When exchanging assertions for access
      tokens, the authorization for the token has been previously
      granted through some out-of-band mechanism.  As such, the
      requested scope MUST be equal to or less than the scope originally
      granted to the authorized accessor.  The authorization server MUST
      limit the scope of the issued access token to be equal to or less
      than the scope originally granted to the authorized accessor.
```
So I think this means a couple things:
 1. Because this is an optional request body parameter per the RFC, and there were no tests for the current gem behavior of it always being there, I'm going to remove this behavior and only put the parameter in if there is a passed-in value.
 2. This means that what I _thought_ scope was doing (getting put into the header of the JWT) was not actually happening . . . that means there is in fact _no_ current functionality that lets you put stuff into the JWT header.  In fact, if you look at the jwt gem's `encode` arguments: 
```ruby
def encode(payload, key, algorithm = 'HS256', header_fields = {})
```
Assuming those arguments are named in a sane manner . . . we've only ever passed in three arguments to that method.  I don't believe we've ever had a way to pass in header fields!

And so while that is interesting and I believe should also be added, I'd like to keep this PR from getting even larger -- and I believe 'adding JWT header field functionality' would be a nice compartmentalized followup PR to make once I'm done with this one.

pboling:
Then with your changes, it will still add the `scope` param to the URL when `scope` is passed in params, but otherwise not add it to the URL, right?

That sounds logical to me, and appears to be in line with that optional RFC. 👍

pboling:
@jhmoore Does/Will your PR obviate the need for this other JWT related PR, which also change `prm` to `sub`?  https://github.com/oauth-xx/oauth2/pull/198/files

jhmoore:
Aha!  Yes, although mine goes about it in a slightly different manner.  That one passes through `params` straight through to the claimset (and certain keys like `hmac_secret` or `scope` are 'special'), whereas this passes `params[:claims]` to the claimset.  Now that I've gotten really into this: most of the fundamental incompatibility this file has with Google is the fact that they require a claim named `scope` -- which directly collides with the concept of `params[:scope]` here.

While we're on the topic: if this is going to be a `2.0.0` breaking change release . . . I'm less opposed to breaking the signature of this method's arguments.  Simply adding another argument to `get_token` would certainly be cleaner than doing a nested hash.  What do you think?

pboling:
Yes, let's go with the breaking change and the better design. 👍

pboling:
This will fix https://github.com/oauth-xx/oauth2/issues/203

pboling:
@jhmoore If you have some time please also [review this PR](https://github.com/oauth-xx/oauth2/pull/166).  I think it is addressing some of the same things.  I'm interested to know if your work here will allow us to close that (stale, conflicted) PR.

jhmoore:
@pboling yep you are correct, you can close that one.

pboling:
@jhmoore One more!  https://github.com/oauth-xx/oauth2/pull/198/files

Does that one also get obviated by this PR?

jhmoore:
@pboling :+1: indeed, that one too!

jhmoore:
@pboling take a look -- there's still some cleanup I'd like to do, and I need to do a full rewrite of the description of this PR, but https://github.com/oauth-xx/oauth2/pull/220/commits/34c37814f68810423eba7cbe1598a70037ff7b36 is effectively what I'm proposing in terms of argument structure changes to `get_token`.  I'm not the biggest fan of having two consecutive `opts = {}` arguments, but the way that those things are used are so drastically conceptually different, I didn't really want to combine them.

I'd also appreciate a review of the documentation comment changes -- I attempted to emulate the style in other files, but haha I've never used this YARD thing before so I have no idea how successful I was.

pboling:
This is also my first time on the YARD.  I'll review later tonight.

pboling:
Just need to get the build passing before we can merge.

jhmoore:
Yeah, was just looking at that.  Appears to be entirely Rubocop . . . . including some rules I disagree with on a deep philosophical level (e.g. `Do not use expect in before hook`) but haha this is not my project.  Will work on that now.

jhmoore:
Oh, boy.  Haha one of the rules is `RSpec/NestedGroups: Maximum example group nesting exceeded [5/3]`.  I can alter the specs, but it's going to make the organization _much_ worse.  Given the specs look good to you, is it alright if I disable that rule?

Also: both Rubocop and Code Climate are angry with the new method having 5 arguments.

jhmoore:
Oh this is interesting: one of the Travis builds is failing for reals.  https://travis-ci.org/oauth-xx/oauth2/jobs/351559454

Specifically, `jruby-1.7.27` is blowing up and unable to call `JWT::Decode` in the specs.  The other versions of jruby appear to be fine.  Haha I know next to nothing about jruby version minutiae . . . any thoughts?

jhmoore:
It appears to be some issue with referencing `JWT::Decode` directly in the spec?  Curiously, the call to `JWT.decode` seems to work fine.  

The question of when `require`-ing a module in a spec, why one specific version of jruby cannot access a class in that module's namespace (despite other versions of jruby being able to access that class just fine) is a keening maw of madness I don't really feel like spelunking into on this particular Friday afternoon, so I just changed the specs to all use `JWT.decode` instead.

pboling:
Yeah, let's disable that RSpec Rubocop validation.  `RSpec/NestedGroups: Maximum example group nesting exceeded [5/3]`.  I tend to like highly contextualized specs as well.  I'd love to be able to have simple objects to test that only do one thing with little context required... but the world is not that simple, and getting things to be that simple is... complicated. 💃

jhmoore:
@pboling alrighty, take a look.  I am not wild about the format of `get_token(hash, hash, hash, hash)`, but Rubocop and Codeclimate both were rather insistent about a maximum of 4 arguments in a method.  

This method _is_ taking in a lot of different conceptual things.  One option might be to have `Assertion` itself be initialized with the claims and encoding options -- i.e. the concept of "an assertion" seems like a thing that would be initialized with its JWT ready to go, and "get token" seems like an instruction that would take that assertion and send it to be exchanged for a token?  However, `Assertion` inherits from `Strategy::Base` which establishes a shared bit of superclass instance magic . . . no other strategy type seems to override `def initialize`.  

But yeah, `Client#assertion` could be modified to take in some args and pass them to the initialized strategy object; usage in that case would be more like:
```ruby
client.assertion(claims, 'HS256', 'secret_key').get_token(opts)
```
Anyway, this PR as it stands right now will work (and I've updated the PR description for the current state of things), but I am open to suggestions.  Haha I guess it depends how closely-shorn we want to groom this particular yak.

.gitignore
```diff
@@ -27,6 +27,7 @@ Gemfile.lock
 .rvmrc
 
 # Editors
+.idea
```
pboling: I am partial to IDEs and those who use them. 💯

.rubocop.yml
```diff
@@ -44,6 +44,9 @@ Lint/UnusedBlockArgument:
     - 'vendor/**/*'
     - '**/.irbrc'
 
+RSpec/NestedGroups:
+  Enabled: false
```
pboling: I approve.  I think this rule is not well suited to the current state of this project.

lib/oauth2/strategy/assertion.rb
```diff
@@ -4,19 +4,21 @@ module OAuth2
...
-    #             :exp => Time.now.utc.to_i + 3600}
+    #   claimset = {
+    #     :iss => "http://localhost:3001",
+    #     :aud => "http://localhost:8080/oauth2/token"
+    #     :sub => "me@example.com",
+    #     :exp => Time.now.utc.to_i + 3600
+    #   }
     #
-    #   access = client.assertion.get_token(params)
+    #   access = client.assertion.get_token(claimset, :algorithm => 'HS256', :key => 'secret_key')
```
pboling: I would prefer to have the second hash param be explicitly surrounded with curly braces, or setup as an instance variable like claimset.

This is because the presumed hash args without explicit curly braces have ways of badly breaking if the claimset hash gets inlined, or other parts become variables.

It isn't immediately clear from the documentation as is that this will not work:
```
algo = {:algorithm => 'HS256'}
key = {:key => 'secret_key'}
access = client.assertion.get_token(claimset, algo, key)
=> Boom!
```

Once we drop old Rubies we can avoid all this ambiguity and move to named parameters!

For now:
```
access = client.assertion.get_token(claimset, { :algorithm => 'HS256', :key => 'secret_key' })
```

pboling:
Approved, but would like this line of documentation tweaked: https://github.com/oauth-xx/oauth2/pull/220#discussion_r174033859

jhmoore:
@pboling agreed . . . when a method has multiple hash args and one of them is optional, I agree that I kinda always want curlies regardless of whether or not I'm passing in the optional one.  Hilariously --
 being a believer of the whole 'tests as documentation' thing, I _did_ have it like that, along with the test setup.  However, rubocop blows up with:
```
Offenses:

spec/oauth2/strategy/assertion_spec.rb:69:37: C: Style/BracesAroundHashParameters: Redundant curly braces around a hash parameter.
          subject.get_token(claims, {:algorithm => algorithm, :key => key})
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Militant elegance at the expense of clarity!  :rage1: Haha anyway, tests are left in the state of rubocop passing; comment is updated as you suggested.

xtagon:
Thank you @pboling and @jhmoore - I know this was a big one!

jhmoore:
👍Thanks for reviving this gem!  

It might take me a bit, but now that this stuff is merged and the changes solidified, I'll work on documentation of the Google service account use case.

pboling:
That would be most excellent!

PR #223, opened 2015-12-09
Dupe options hash to prevent modification of original credentials hash

jwg2s:
Pretty self-explanatory - don't leave it up to chance that the user is passing a hash into the gem that is ok to be mutated.


sferik:
Dupe of https://github.com/intridea/oauth2/pull/240. 😉

PR #230, opened 2016-01-14
Update oauth2.gemspec

donaldpiret:
Change the dependency to rack to be less than 3 in order to work with rails 5


donaldpiret:
There seem to be some PR's open for this already

PR #236, opened 2016-01-27
Add jwt bearer and saml assertions flows

usbsnowcrash:
This PR adds two more strategies (SAML and JWT-Bearer)


usbsnowcrash:
Looking at travis failures

usbsnowcrash:
RBX failed?

PR #237, opened 2016-01-28
Test support for rack 2.0 on Travis

pschambacher:
@sferik 

Hi!

This gem is on our path to get our gems compatible with Rails 5.0.0.beta1. I noticed that you unlocked the hard dependency on `rack` but it wasn't exactly tested yet. This should do.

I also unlocked the dependency on jwt since it's only for ruby 1.8.

Would you mind releasing a new version, at least alpha/beta, so we can bundle with rails 5 and this gem?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 .bundle
 .rvmrc
 Gemfile.lock
+Gemfile.rack2.lock
 coverage/*
 log/*
 measurement/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,11 @@ rvm:
   - rbx-2
   - ruby-head
 matrix:
+  include:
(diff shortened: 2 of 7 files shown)


shinnn:
:+1:

pschambacher:
:eyes: @sferik ?

sferik:
I already released the change that allows `rack` version 2 and relaxed the `jwt` dependency in the gemspec in `master`. Does Rails require a newer version of `jwt`?

pschambacher:
It doesn't. This was just to get the tests running on both versions to be sure because the test suite is only running rack 1.x at the moment.

sferik:
Closed by https://github.com/intridea/oauth2/commit/22d40c7045224f604c1e4224f5e568a4c7be858b.

PR #240, opened 2016-02-05
dup the options hash to avoid accidental mutation

### LINKED ISSUE #215: AcessToken #from_hash and #new delete from the hash
if you want to own the passed hash it should be duped

Linuus:
Fixes #215 

Hi!

I just had an annoying bug with accidental mutation of a hash.

```ruby
class Client
  def initialize(token, opts = {})
    @token = token
    @opts = opts
  end

  def access_token
    OAuth2::AccessToken.new(client, @token, @opts)
  end
   
  def client
    # ...
  end
end

client = Client.new("123123123", refresh_token: "refresh-token-123", expires_at: "123123123")

client.access_token.refresh_token #=> "refresh-token-123"

# Run it again
client.access_token.refresh_token #=> nil
```

Why does this happen?
Here: https://github.com/intridea/oauth2/blob/master/lib/oauth2/access_token.rb#L43
We run `opts.delete(...)` which changes the opts hash in-place.

Instead, we should `dup` the hash to avoid this issue.

Any input?

### DIFF
--- a/lib/oauth2/access_token.rb
+++ b/lib/oauth2/access_token.rb
@@ -10,6 +10,7 @@ module OAuth2
       # @param [Hash] a hash of AccessToken property values
       # @return [AccessToken] the initalized AccessToken
       def from_hash(client, hash)
+        hash = hash.dup
         new(client, hash.delete('access_token') || hash.delete(:access_token), hash)
       end
 
@@ -39,6 +40,7 @@ module OAuth2
     def initialize(client, token, opts = {}) # rubocop:disable Metrics/AbcSize
       @client = client
       @token = token.to_s
+      opts = opts.dup
       [:refresh_token, :expires_in, :expires_at].each do |arg|
         instance_variable_set("@#{arg}", opts.delete(arg) || opts.delete(arg.to_s))
       end
--- a/spec/oauth2/access_token_spec.rb
+++ b/spec/oauth2/access_token_spec.rb
@@ -44,6 +44,13 @@ describe AccessToken do
       assert_initialized_token(target)
     end
 
+    it 'does not modify opts hash' do
+      hash = {:access_token => token, :expires_at => Time.now.to_i}
+      hash_before = hash.dup
+      AccessToken.from_hash(client, hash)
+      expect(hash).to eq(hash_before)
+    end
+
     it 'initalizes with a form-urlencoded key/value string' do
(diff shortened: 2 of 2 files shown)


Linuus:
@sferik Any chance of having this reviewed/merged? It shouldn't affect any current functionality.

JoshCheek:
Toss a test on it?

Linuus:
Done!

lib/oauth2/access_token.rb
```diff
@@ -10,6 +10,7 @@ class << self
       # @param [Hash] a hash of AccessToken property values
       # @return [AccessToken] the initalized AccessToken
       def from_hash(client, hash)
+        hash = hash.dup
```
JoshCheek: This one seems redundant, looks like the `dup` added to initialize will already get this.

spec/oauth2/access_token_spec.rb
```diff
@@ -44,6 +44,13 @@ def assert_initialized_token(target) # rubocop:disable Metrics/AbcSize
       assert_initialized_token(target)
     end
 
+    it 'does not modify opts hash' do
+      hash = {:access_token => token, :expires_at => Time.now.to_i + 200, 'foo' => 'bar'}
+      hash_before = hash.dup
+      AccessToken.from_hash(client, hash)
+      expect(hash).to eq(hash_before)
+    end
+
```
JoshCheek: This test is probably unnecessary for the same reason.

lib/oauth2/access_token.rb
JoshCheek: Nvm, I'm wrong, it's relevant for the access token on the next line.

spec/oauth2/access_token_spec.rb
JoshCheek: I'm wrong here, too: `AccessToken.from_hash` deletes the access token, so this does show something relevant. The `'foo' => 'bar'` is probably not relevant. The `:expires_at => Time.now.to_i+200` shows that `#initialize` doesn't modify it, though it could be a red-herring to have the value look so intentionally chosen.

Linuus:
@JoshCheek I fixed your comments. I don't know why Travis seems to fail though... It can't even install rake?

sferik:
@Linuus Thanks for the patch!

@JoshCheek Thanks for reviewing!

PR #243, opened 2016-03-10
[Microsoft] fix reading OpenID Connect token responses

ccutrer:
when using Microsoft's OpenID Connect service (as documented at
https://azure.microsoft.com/en-us/documentation/articles/active-directory-v2-protocols/)
if you only request the openid scope, Microsoft only returns an id_token,
with no access_token. so treat that as a valid response.

### DIFF
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -130,7 +130,7 @@ module OAuth2
     # @param [Hash] access token options, to pass to the AccessToken object
     # @param [Class] class of access token for easier subclassing OAuth2::AccessToken
     # @return [AccessToken] the initialized AccessToken
-    def get_token(params, access_token_opts = {}, access_token_class = AccessToken) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+    def get_token(params, access_token_opts = {}, access_token_class = AccessToken) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength, Metrics/PerceivedComplexity
       params = authenticator.apply(params)
       opts = {:raise_errors => options[:raise_errors], :parse => params.delete(:parse)}
       headers = params.delete(:headers) || {}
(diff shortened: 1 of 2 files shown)


lib/oauth2/client.rb
```diff
@@ -136,8 +136,11 @@ def get_token(params, access_token_opts = {}, access_token_class = AccessToken)
         opts[:params] = params
       end
       response = request(options[:token_method], token_url, opts)
-      error = Error.new(response)
-      raise(error) if options[:raise_errors] && !(response.parsed.is_a?(Hash) && response.parsed['access_token'])
+      if options[:raise_errors] && !(response.parsed.is_a?(Hash) &&
```
jcarbo: This line is a little tough to read. I'd change to something like this:

```ruby
response_contains_token = response.parsed.is_a?(Hash) &&
  (response.parsed['access_token'] || response.parsed['id_token'])

error = Error.new(response)
raise(error) if options[:raise_errors] && !(response_contains_token)
```

Also this method now breaks the lint check, which is failing the build. In a method above there's a comment to disable a few extra Rubocops, I'd add the same comment to this method `def` as well:

```ruby
# rubocop:disable CyclomaticComplexity, MethodLength, Metrics/AbcSize
```

jcarbo:
I've run into the same issue. I left some comments to help fix the build and then hopefully this could get some attention from 👍

ccutrer:
I've updated the PR with your suggestions.

xtagon:
@ccutrer Are you still interested in getting this merged? Looks like there are just some conflicts now.

natebird:
Any movement on getting this merged @ccutrer?

ccutrer:
I'd almost forgotten this PR existed. I've just been updating our monkey patch in our app with each bump of this gem. I'll update the PR

ccutrer:
PR updated with conflicts resolved, and specs passing.

jonspalmer:
I know this is an old PR but is there any advice on how to use Microsoft responses that only hav id_token in the. This fix make that a valid response but the AccessToken class still expects the `access_token` in the response hash.

What's the fix? Is it to subclass AccessToken ?

anvox:
Thansk @jonspalmer I afraid extending AccessToken is the only option for now. I see you found https://github.com/oauth-xx/oauth2/pull/511 already, we could move conversation on it for a completely solution for this.

PR #252, opened 2016-05-12
Add support to ClientCredentials to support authorization formats other than just HTTP Basic

brycejohnston:
Adds support to ClientCredentials get_token and authorization methods to support authorization formats other than just HTTP Basic Authentication, such as, Bearer Authentication. Defaults to 'Basic' if header_format is not defined in opts hash of get_token method.

``` ruby
# Basic auth works the exact same as before
@new_token = @client.client_credentials.get_token({
  'client_id' => client_id,
  'client_secret' => client_secret})

# Bearer auth, pass in hash as third param with :header_format defined
@new_token = @client.client_credentials.get_token({
  'client_id' => client_id,
  'client_secret' => client_secret},
  {:header_format' => 'Bearer'})
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,8 @@ token = OAuth2::AccessToken.from_kvform(client, query_string)
 
 token = client.password.get_token('username', 'password')
 
-token = client.client_credentials.get_token
+# :header_format - the string format to use for the Authorization header
+token = client.client_credentials.get_token({'client_id' => client_id, 'client_secret' => client_secret}, {:header_format' => 'Basic'})
 
 token = client.assertion.get_token(assertion_params)
 ```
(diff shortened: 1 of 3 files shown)


brycejohnston:
@sferik Not sure what is happening, looks like the Travis CI build is not attempting to run on this pull request due to some kind of error?

brycejohnston:
@sferik I know the [OAuth 2 specification in section 2.3.1](https://tools.ietf.org/html/rfc6749#section-2.3.1) says that the client credentials flow must support the HTTP Basic authentication scheme, but this pull request allows those working with APIs that aren't fully compliant and require something else, like Bearer, to be able to specify an alternative authentication scheme.

Any possibility of this getting merged?

niedakh:
is there a specific reason why this is not merged? I need this very badly

PR #251, opened 2016-05-01
Fix encoding error

Motoshi-Nishihira:
I want to fix #155 

It will occur when Content-Type is `application/json` and response.body's includes non ascii characters.
I think the correct encoding is dependent on the script encoding(`__ENCODING__`).

PR #253, opened 2016-05-13
Remove dependency on encoding class

Motoshi-Nishihira:
fix broken build about #251

Encoding class are implemented as of ruby 1.9.

In ruby 1.8, string don't have the encoding information.
So no error is raised by joining different encoding strings.
```
irb(main):001:0> ["∞".kconv(Kconv::EUC, Kconv::UTF8), "∞"].join
=> "\241\347\342\210\236"
```

As of 1.9
```
irb(main):001:0> ["∞".kconv(Kconv::EUC, Kconv::UTF8), "∞"].join
Encoding::CompatibilityError: incompatible character encodings: EUC-JP and UTF-8
```

### DIFF
--- a/lib/oauth2/error.rb
+++ b/lib/oauth2/error.rb
@@ -1,5 +1,3 @@
-require 'kconv'
-
 module OAuth2
   class Error < StandardError
     attr_reader :response, :code, :description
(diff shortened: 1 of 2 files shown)


sferik:
It looks like some of our code quality metrics are now broken. Would you mind refactoring this pull request to fix those?

sferik:
Thanks!

PR #255, opened 2016-06-10
Parse 'application/xml' content type as XML in OAuth2::Response

akowalz:
I am working with an OAuth2 API that serializes OAuth credentials (`access_token`, `refresh_token`, etc) as XML with the content type`application/xml`.  This content type should be registered to parse as XML by the `OAuth2::Response` class, but is currently absent from the list of content types to be parsed as XML.  This commit simply adds `application/xml` to this list.

### DIFF
--- a/lib/oauth2/response.rb
+++ b/lib/oauth2/response.rb
@@ -84,6 +84,6 @@ module OAuth2
   end
 end
 
-OAuth2::Response.register_parser(:xml, ['text/xml', 'application/rss+xml', 'application/rdf+xml', 'application/atom+xml']) do |body|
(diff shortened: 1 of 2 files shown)


josephpage:
Thanks @akowalz for your contribution !
Can you please add a spec for this case ?
An example is the spec for `text/xml` in [response_spec.rb#L83](https://github.com/oauth-xx/oauth2/blob/da9c7ad8b01d5a464bd3005645718a69a9fb9d1a/spec/oauth2/response_spec.rb#L83).

akowalz:
@josephpage Spec added!

pboling:
Thanks @akowalz

PR #256, opened 2016-06-17
handle gzip encoded content

ChaosSteffen:
Basically this is an updated version of this pull request: https://github.com/intridea/oauth2/pull/159

We are still having issues with this in 2.2.5, which causes us to do some very dirty hacks around that.

### DIFF
--- a/lib/oauth2/response.rb
+++ b/lib/oauth2/response.rb
@@ -60,7 +60,7 @@ module OAuth2
 
     # The HTTP response body
     def body
-      response.body || ''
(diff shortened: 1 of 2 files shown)


josephpage:
As `oauth2` does not send `Accept-Encoding: gzip` header in the request, the HTTP(S) server must not respond with a compressed body.
To support gzip compression (which is a brilliant idea !) both request and response should work together. 
For me using the [gzip Faraday middleware](https://github.com/lostisland/faraday_middleware/wiki/Gzip-Compression) is a better way, as it supports deflate and brotli too.

PR #267, opened 2016-08-08
fix(encoding): Handle invalid encoding 

fjg:
This is due to Faraday response.body which is set to ASCII-8BIT instead of (request header charset, i.e. UTF-8). 
Seams related to a underlying adapter: see https://github.com/lostisland/faraday/issues/139

### DIFF
--- a/lib/oauth2/error.rb
+++ b/lib/oauth2/error.rb
@@ -27,7 +27,7 @@ module OAuth2
 
       error_message = if opts[:error_description] && opts[:error_description].respond_to?(:encoding)
                         script_encoding = opts[:error_description].encoding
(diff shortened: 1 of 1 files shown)


ArnoHolo:
CI returns the following errors with Ruby 1.8.7 and 1.9.3 :  
> Gem::InstallError: json requires Ruby version ~> 2.0.
> An error occurred while installing json (2.0.2), and Bundler cannot continue.

Is there a way to fix this ? I don't see why this PR would brought this error...

sarojkh:
The issue this PR is trying to solve seems to have been already solved by https://github.com/oauth-xx/oauth2/pull/380.

pboling:
@sarojkh Thanks for reviewing!

Issue #265, opened 2016-08-01
redirect uri incompatible with Doorkeeper


SophieDeBenedetto:
There seems to be an issue when building out a custom OmniAuth strategy to be used when authenticating with an application that uses Doorkeeper. 

After first request for authorization request is completed and authorization code is returned, the redirect URI has that code and the status appended to it. 

![oauth-note](https://cloud.githubusercontent.com/assets/8999596/17306212/d0238b12-57fc-11e6-8674-ca798f86cf2c.png)

When the subsequent request to the token URL is made, Doorkeeper seems to be unable to recognize the altered redirect URI as valid and the following error is thrown: 

![screen shot 2016-08-01 at 3 22 12 pm](https://cloud.githubusercontent.com/assets/8999596/17306249/f9bcbeb2-57fc-11e6-9777-0e3219698c80.png)

I don't think that this is an issue with my custom OmniAuth strategy. You can find the source for that strategy, bundled as a gem [here](https://github.com/flatiron-labs/omniauth-learn) and for a quick look at how that strategy is defined:

``` ruby
require 'omniauth-oauth2'

module OmniAuth
  module Strategies
    class Learn < OmniAuth::Strategies::OAuth2
      include OmniAuth::Strategy

      option :client_options, {
               site: "https://learn.co",
               authorize_url: "https://learn.co/oauth/authorize",
               token_url: "https://learn.co/oauth/token"
             }

      def request_phase
        super
      end

      info do
        raw_info.merge("token" => access_token.token)
      end

      uid { raw_info["id"] }

      def raw_info
        @raw_info ||= access_token.get('/api/users/me').parsed
      end
    end
  end
end
```

By checking to see if the request being made was the secondary request to the token URL, and changing the redirect URI to remove the appended authorization code and status if so, has fixed the problem for me. I created a fork of OAuth2 and my fix can be found [here](https://github.com/flatiron-school/oauth2/commit/bf389481ec02f65aa2885a787dbbad342af084a3).

I did some googling, and found a number of other people who encountered the same issue when building a custom OmniAuth strategy, authorizing with a service that uses Doorkeeper to handle OAuth. 

If this is in fact an incompatibility with Doorkeeper, I would recommend incorporating this change into OAuth2, as Doorkeeper is a fairly common library for handling OAuth in Rails.

dankozlowski:
I've run in to this issue as well, any hope in pushing that PR forward?

COConsulting:
Is this still an active request?

[...]

PR #269, opened 2016-08-25
fix for Doorkeeper incompatibility

SophieDeBenedetto:
Fixes issue #265 

When secondary authorization request is made to the token url, OAuth2 appends the status code the to redirect URI. This is incompatible with Doorkeeper which then refuses to recognize the registered callback url. I made a minor change to `client.rb` at LOC 92-98--by checking to see if the request being made was the secondary request to the token URL, and changing the redirect URI to remove the appended authorization code and status if so.

### DIFF
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -89,8 +89,14 @@ module OAuth2
       connection.response :logger, ::Logger.new($stdout) if ENV['OAUTH_DEBUG'] == 'true'
 
       url = connection.build_url(url, opts[:params]).to_s
-
-      response = connection.run_request(verb, url, opts[:body], opts[:headers]) do |req|
+      body = nil
+      if opts[:body]
+        opts[:body][:redirect_uri] = opts[:body][:redirect_uri].split("?").first
+        body = URI.encode_www_form(opts[:body])
+      else
+        body = opts
+      end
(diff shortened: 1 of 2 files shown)


tswayne:
Looks like this pull request will depend on https://github.com/intridea/oauth2/pull/263 to be merged in order for the build to run successfully.

neerfri:
I think the problem you are referring to is in the `omniauth-oauth2` gem.
See: https://github.com/intridea/omniauth-oauth2/issues/93
A workaround is given at: https://github.com/intridea/omniauth-oauth2/issues/93#issuecomment-227247801
The PR that causes the issue: See https://github.com/intridea/omniauth-oauth2/pull/70/files

dankozlowski:
Any way we could push this fix over the finish line?

johaywood:
@neerfri is correct. this isn't an issue with this gem, but rather with omniauth-oauth2. I've used the fix mentioned above in many omniauth strategies.

pboling:
Thanks for the fix @neerfri & for the analysis and bump @johaywood .

Any interest in taking on a broader maintenance role in this project?

johaywood:
@pboling I'd be interested in helping out with maintenance. I've certainly relied on this project heavily, and continue to, so I definitely don't want to see it fall by the wayside. Let me know where the priorities are and I'll see where I can jump in.

Issue #270, opened 2016-09-21
Encoding::UndefinedConversionError while parsing responses.


urkle:
Asana returns extra data in the get_token reponse pertaining to the current user (which MAY have unicode UTF-8 data).  Because the underlying network layer (ruby's Net::HTTP) returns data as ASCII-8BIT, the current code causes this error to occur.

Note this is a SUCCESSFUL token reponse (e.g. no errors)

Backtrace

```
/gems/oauth2-1.2.0/lib/oauth2/error.rb:30 in encode
/gems/oauth2-1.2.0/lib/oauth2/error.rb:30 in error_message
/gems/oauth2-1.2.0/lib/oauth2/error.rb:17 in initialize
/gems/oauth2-1.2.0/lib/oauth2/client.rb:139 in new
/gems/oauth2-1.2.0/lib/oauth2/client.rb:139 in get_token
/gems/oauth2-1.2.0/lib/oauth2/strategy/auth_code.rb:29 in get_token
```

A few questions.
- Why is an error instance always created on line 139 of client.rb if the check for having an error in the first place happens on line 140?  It would make better sense to create that instance IN the raise(error) instead?
- why is the error message routine changing the encoding of the response body anyways? and why to the "hand-built" string up on line 14?
  - which is always US-ASCII in this scenario as code and description are both nil.
    - "#{nil}: #{nil}".encoding == US-ASCII
    - outputting a nil in that above string cause the encoding to degenerate to US-ASCII, even if one is UTF-8 or ASCII-8BIT

However, regardless, even if the error string data WERE correct, if there is ANY UTF-8 data in the string, the encode line will cause an error.

``` ruby
body = 'sometext with UTF-8 André '.force_encoding('ASCII-8BIT')
body.encode('US-ASCII')  # will raise Encoding::UndefinedConversionError:"\xC3" to UTF-8 in conversion from ASCII-8BIT to UTF-8 to US-ASCII
body.encode('UTF-8')  # will raise Encoding::UndefinedConversionError: "\xC3" from ASCII-8BIT to UTF-8
body.encode('ASCII-8BIT') # raises nothing
```

There are a few things that I see need to be changed here.
1. the error class should not be instantiated if there is no error detected (thus avoiding 99% of this nonsense)
2. The re-encoding should simply not happen, OR it should have a "replace" mode activated e.g. `response_body.encode(script_encoding, :invalid => :replace, :undef => :replace)`

[...]

PR #271, opened 2016-09-21
Fix error encoding

urkle:
Fix for #270 

This makes two changes
- adds handling of unknown encoding bytes when re-encoding the error response
- only instantiates the Error class if an actual error has occurred.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ rvm:
   - ruby-head
 matrix:
   allow_failures:
+    - rvm: 1.8.7
     - rvm: jruby-head
     - rvm: ruby-head
   fast_finish: true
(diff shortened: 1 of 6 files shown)


urkle:
@sferik I fixed the rubycop issues in this.. 

I also tweaks the Gmefile and .travis.yml as ruby 1.8.7 no longer tests correctly, and 1.9.3 requires restricting gem versions to pass.

If you could please review this and provide feedback, that would be awesome, as this has caused significant issues in our production environments.

robertIngrum:
👍 Would be great to get this out, causing significant issues for us as well.

robertIngrum:
@sferik: Just pinging again, would love to see this on master.

jcarbo:
Is there anything holding this up? We're using this gem with Office365 which returns non-ascii in some cases and our integration is currently failing for some of our customers. Would appreciate this making it into a release 👍

jwg2s:
Anything we can do to help get this merged?

robertIngrum:
I think @sferik is the only admin, so just ping and hope it gets noticed I guess.

sferik:
Thanks for this fix!

jwg2s:
@sferik worth issuing a v1.2.1 release for this?  Pretty bad bug

PR #276, opened 2016-11-18
Loosen faraday requirements to allow 0.10 to be picked up

andrew:
Used a similar upper version to what @mislav did in https://github.com/lostisland/faraday_middleware/commit/6532e65c3e25ce8bdf884d118602a41efc0d42e3#diff-2c97a3fba04553020532f4508b0b3bb4


sferik:
Resolved by https://github.com/intridea/oauth2/pull/277.

PR #277, opened 2016-11-30
Add support for Faraday 0.10

rhymes:
Fixes https://github.com/intridea/oauth2/issues/275

### DIFF
--- a/oauth2.gemspec
+++ b/oauth2.gemspec
@@ -4,7 +4,7 @@ $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 require 'oauth2/version'
 
 Gem::Specification.new do |spec|
-  spec.add_dependency 'faraday', ['>= 0.8', '~> 0.10']
+  spec.add_dependency 'faraday', ['>= 0.8', '< 0.11']
   spec.add_dependency 'jwt', '~> 1.0'
   spec.add_dependency 'multi_json', '~> 1.3'
   spec.add_dependency 'multi_xml', '~> 0.5'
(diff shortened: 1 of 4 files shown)


chulkilee:
I'm not maintainer, but I think it would be nice to include only related commit (e.g. no rubocop fixes) in ths PR.

oauth2.gemspec
```diff
@@ -4,7 +4,7 @@ $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 require 'oauth2/version'
 
 Gem::Specification.new do |spec|
-  spec.add_dependency 'faraday', ['>= 0.8', '< 0.10']
+  spec.add_dependency 'faraday', ['>= 0.8', '~> 0.10']
```
chulkilee: For consistency, why don't you just use `< 0.11`?

rhymes:
@chulkilee you're right but the build does not pass (and hence can't be merged) without the rubocop fixes, that's why they are in there.

oauth2.gemspec
chulkilee: First of all, that's different from what the title of this PR says :)

Also you cannot do that since faraday does not promise backward compatibility. https://github.com/lostisland/faraday/blob/master/CONTRIBUTING.md says it will follow semver *once* it hits 1.0. So you should use `< 0.11`.

rhymes: @chulkilee done

athix:
I need this fix to resolve dependency mismatch between this and the twitter gem I'm using. +1 👍

PR #278, opened 2016-11-30
Bump rack version in Gemfile to ~> 2.0

tiabas:
The version of Rack in the Gemfile is `~> 1.2` whereas in the gemspec it is `[>=1.0, <3]`. For projects relying on this gem but requiring `rack ~> 2.0` bundler fails because it builds agains thte definition in the Gemfile. I have bumped the version so that projects requiring a higher version of rack can still build their bundles

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,7 +13,7 @@ group :test do
   gem 'backports'
   gem 'coveralls'
   gem 'mime-types', '~> 1.25', :platforms => [:jruby_18, :ruby_18]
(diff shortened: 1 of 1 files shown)


tiabas:
May I get some input on how best I can make this change. I'm trying to update the gem version so that I can get Omniauth gem to build on Rails 5.0.0.1. The minimum rack version in Rails 5 is `2.0`

josephpage:
Thanks @tiabas 
Probably already fixed in #326.

pboling:
@josephpage I believe #326 kept rack at 1.2.0 in the Gemfile.  We still need to address this, though I'm not sure of the best way to do that, or if this PR is exactly what we want.  There may be other considerations in a major version bump of rack.

PR #280, opened 2016-12-01
Allow authenticating with Basic Auth for token exchange

bjeanes:
This is a re-work of #192.

For now, it builds on that PR to maintain credit where it is due to @maletor, though I am not opposed to rebasing and re-arranging the narrative here to make it clearer for reviewers.

This takes a slightly different approach than #192 in a few areas:
 
* The authentication logic is no longer in the `Base` strategy so that it is accessible to token refreshes as well (though, alternatively, refreshing could be made a `Strategy` too...)
* Auth logic is in its own class (both because Rubocop was warning against the length of `Client` and because this makes testing some edge cases straight forward)
* This aims to be a non-breaking change. It does _not_ change the default authentication method (though I do think Basic Auth is a better default for a future, breaking version) so perhaps a note in any release notes is appropriate
* This will not clobber existing params/headers when applying authentication.

There is a few open questions which I'd like input on and then I can quickly address:

* Backwards-compatibility for people passing `auth_scheme` option directly to `OAuth2::Strategy::ClientCredentials#get_token` needs some consideration.
* I'd like to rename the options from `:request_body` to `:request_params` because strictly speaking I do not believe the OAuth 2.0 spec requires these requests to be `POST` so better to have agnostic terms here, but this impacts the backwards-compatibility point above.
* This does **NOT** fix the existing issue of not fully form-encoding the credentials (see [this comment](https://github.com/intridea/oauth2/pull/192#r17500191)). I initially did this but in testing with several providers, credentials were rejected. I think this should be tackled with a separate line of work to keep this PR focussed.

cc @sferik @ozgg @maletor; I'm willing to be quite responsive over the next week to get this over the line so if you want me to take this in a different direction (e.g. in-lining the auth stuff into `Client`) just let me know.

### DIFF
--- a/lib/oauth2.rb
+++ b/lib/oauth2.rb
@@ -1,4 +1,5 @@
 require 'oauth2/error'
+require 'oauth2/authenticator'
 require 'oauth2/client'
 require 'oauth2/strategy/base'
 require 'oauth2/strategy/auth_code'
--- a/lib/oauth2/access_token.rb
+++ b/lib/oauth2/access_token.rb
@@ -81,8 +81,6 @@ module OAuth2
     # @note options should be carried over to the new AccessToken
     def refresh!(params = {})
       raise('A refresh_token is not available') unless refresh_token
-      params[:client_id] = @client.id
-      params[:client_secret] = @client.secret
       params[:grant_type] = 'refresh_token'
       params[:refresh_token] = refresh_token
       new_token = @client.get_token(params)
--- a/lib/oauth2/authenticator.rb
+++ b/lib/oauth2/authenticator.rb
@@ -1,0 +1,58 @@
+require 'base64'
+
+module OAuth2
+  class Authenticator
+    attr_reader :mode, :id, :secret
+
+    def initialize(id, secret, mode)
+      @id = id
+      @secret = secret
+      @mode = mode
+    end
+
+    # Apply the request credentials used to authenticate to the Authorization Server
+    #
+    # Depending on configuration, this might be as request params or as an
+    # Authorization header.
+    #
+    # User-provided params and header take precedence.
+    #
+    # @param [Hash] params a Hash of params for the token endpoint
+    # @return [Hash] params amended with appropriate authentication details
+    def apply(params)
+      case mode.to_sym
+      when :basic_auth
+        apply_basic_auth(params)
+      when :request_body
+        apply_params_auth(params)
+      else
+        raise NotImplementedError
+      end
+    end
+
+    def self.encode_basic_auth(user, password)
+      'Basic ' + Base64.encode64(user + ':' + password).delete("\n")
+    end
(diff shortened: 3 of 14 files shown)


ozgg:
So... will this PR be merged? :) Travis complains only about unset environmental variables.

bjeanes:
@ozgg I hope someone will have the time to review it soon. I'm dedicated to getting this and a few other improvements I have in mind (spec-adherence and API ergonomics) over the line.

ozgg:
Just reminding that people need this PR. @sferik, can you do something for merging it?

sferik:
@bjeanes Thanks for reviving this old PR! I appreciate the thoughtfulness with which you’ve done it, starting with @maletor’s branch and then layering your work on top.

> There is a few open questions which I'd like input on and then I can quickly address:
>
> * Backwards-compatibility for people passing `auth_scheme` option directly to `OAuth2::Strategy::ClientCredentials#get_token` needs some consideration.

This was not a documented option, so I think this is an acceptable change.

> * I'd like to rename the options from `:request_body` to `:request_params` because strictly speaking I do not believe the OAuth 2.0 spec requires these requests to be `POST` so better to have agnostic terms here, but this impacts the backwards-compatibility point above.

Quoting the [OAuth 2 specification](https://tools.ietf.org/html/rfc6749#section-4.1.3) (emphasis mine):

> The client makes a request to the token endpoint by sending the following *parameters*…in the HTTP request entity-*body*

The spec describes the values in question as “parameters” that go in the “body”. I think either name is acceptable, so the name that is less likely to break things is preferred.

> * This does **NOT** fix the existing issue of not fully form-encoding the credentials (see [this comment](https://github.com/intridea/oauth2/pull/192#r17500191)). I initially did this but in testing with several providers, credentials were rejected. I think this should be tackled with a separate line of work to keep this PR focussed.

That’s fine.

bjeanes:
@sferik thanks for the review.

> I think either name is acceptable, so the name that is less likely to break things is preferred.

In that case, let's just leave it as is. Another consideration in this area (whether now or for the future) is that we already have some auth-mode-switching code [in the `AccessToken` class](https://github.com/intridea/oauth2/blob/c9bbb9c58ded9b330ee105778f66f86806335d63/lib/oauth2/access_token.rb#L155-L162). The ergonomics of the gem may benefit from some consistency in naming in those two places.

Regardless, let's assume we leave this param as is, is there anything else I need to do before this can be merged? The tests pass, but the Coveralls and Rubucop and Yardstick output vary from Ruby version and seem to be mostly pre-existing. I can sprinkle some more docs through the code to raise the documentation coverage if you like. Just let me know what the clearest path to a merge for this (and ideally #281) is.

Thanks!

ozgg:
So, @sferik, how are things going with this PR?

PR #281, opened 2016-12-01
Handle redirect_uri according to OAuth 2 spec

bjeanes:
Passing redirect_uri to authorization server at grant time is optional (though encouraged and very commonly enforced by providers). However, if it is passed (as it often is) it is REQUIRED to be passed when using the grant code to get an access token.

This change allows a user to optionally provide a `redirect_uri` when configuring the client to have it passed as a parameter in both the redirect to the authorization server and the later token exchange.

Relevant sections of the spec:

+ https://tools.ietf.org/html/rfc6749#section-4.1
+ https://tools.ietf.org/html/rfc6749#section-4.1.3
+ https://tools.ietf.org/html/rfc6749#section-4.2.1
+ https://tools.ietf.org/html/rfc6749#section-10.6

This change is backwards compatible, its use is optional, and it handles more of the spec so that the user does not have to.

Note: this will conflict with #280. If both are cleared for merging, I'll be happy to rebase the other for a clean merge promptly.

### DIFF
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -183,7 +183,7 @@ module OAuth2
       @assertion ||= OAuth2::Strategy::Assertion.new(self)
     end
 
-    # The redirect_uri parameters, if configured.
+    # The redirect_uri parameters, if configured
     #
     # The redirect_uri query parameter is OPTIONAL (though encouraged) when
     # requesting authorization. If it is provided at authorization time it MUST
@@ -192,10 +192,13 @@ module OAuth2
     # Providing the :redirect_uri to the OAuth2::Client instantiation will take
     # care of managing this.
     #
+    # @api semipublic
+    #
     # @see https://tools.ietf.org/html/rfc6749#section-4.1
     # @see https://tools.ietf.org/html/rfc6749#section-4.1.3
     # @see https://tools.ietf.org/html/rfc6749#section-4.2.1
     # @see https://tools.ietf.org/html/rfc6749#section-10.6
+    # @return [Hash] the params to add to a request or URL
     def redirection_params
       if options[:redirect_uri]
         {'redirect_uri' => options[:redirect_uri]}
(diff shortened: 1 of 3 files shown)


lib/oauth2/client.rb
```diff
@@ -62,7 +63,8 @@ def connection
     # The authorize endpoint URL of the OAuth2 provider
     #
     # @param [Hash] params additional query parameters
-    def authorize_url(params = nil)
+    def authorize_url(params = {})
+      params = (params || {}).merge(redirection_params)
```
bjeanes: I wonder if it is more correct (spec-wise) for these redirection params to only be added to `AuthCode#authorize_params` instead.

sferik:
After merging your other PRs, this one is now in conflict. Please resolve the conflicts and I will merge it.

bjeanes:
@sferik done! The coverage has increased from the threshold specified in the `Rakefile` but otherwise the tests are passing. Do you want me to adjust the threshold or do you want to do that when you merge?

sferik:
@bjeanes Thanks! Don’t worry about adjusting the YARD threshold. I’ll patch that up.

I’m planning to push a release soon, unless there are other changes you want to make before I do.

bjeanes:
@sferik thanks 🎉 

> I’m planning to push a release soon, unless there are other changes you want to make before I do.

There's a chance I'll continue to get involved over next few months. There's a few things I'd like to contribute in a few areas (spec-compliance, ergonomics, architecture) if you're open to it.

The thing I'd like to approach most due to my own use case is extracting `AccessToken`'s authentication logic out to a separate concern. In my case, the authorization server != API server, which mean's I'm reimplementing the authentication logic that's buried inside `AccessToken` and tied to interacting with the hostname used for authorization itself.

Nonetheless, I probably won't tackle that until the new year at the earliest and probably some months into it, so I wouldn't hold up any planned releases for that.

sferik:
@bjeanes That all sounds good to me. I’ll push a release now.

sferik:
Shipped! :shipit:

PR #284, opened 2016-12-30
Allow expires_in value of 0

magicalbanana:
👨 🐻 🐷 

If `expires_in` has a value of `0`, `expires_at` gets set. This shouldn't be the case.

Use case:
VKontakte returns a `0` value for `expires_in` for offline access, see https://vk.com/dev/permissions.


josephpage:
Following the RFC, a `0` value in `expires_in` means the access_token expires as soon as emitted, and not valid forever.

https://tools.ietf.org/html/rfc6749#section-4.2.2
>The lifetime in seconds of the access token.  For
>         example, the value "3600" denotes that the access token will
>         expire in one hour from the time the response was generated.
>         If omitted, the authorization server SHOULD provide the
>         expiration time via other means or document the default value.

pboling:
Agree with @josephpage , this can be closed.

PR #288, opened 2017-01-18
Allow faraday 0.11.x

rhymes:


etiennebarrie:
Lots of Rubocop violations, but 👍  on the idea.
Maybe we could even remove the maximum version altogether?

PR #289, opened 2017-01-19
Relax faraday constraint

phlipper:
This commit adds support for versions 0.10.x and 0.11.x which have been released recently. It will also support 0.x until 1.0, which matches the existing pessimistic dependency version constraints and should work as long as `faraday` adheres to SemVer conventions.


shaneog:
Would be great to get this merged and released. Is there anything I can do to help ?

tinogomes:
bump

PR #296, opened 2017-03-10
Correcting the default value doc for the auth_scheme option

jochenschneider:


josephpage:
`:basic_auth` is now the default (#312). Sorry for the confusion.

Issue #186, opened 2014-07-23
get_token should put Response headers and status in the AccessToken object


veetow:
It would be very useful for debugging to have the `Response.header` and `Response.status` parameters merged into the returned `AccessToken` object as there's no way to bubble that full `Response` object up when calling the `get_token` method.

jskirst:
In addition to debugging, I've run into a situation where the response contains a header value that is necessary for further requests.

xtagon:
Anyone up for doing a PR?

pboling:
Resolved by #298

PR #298, opened 2017-03-24
Set the response object on the access token on Client#get_token

cpetschnig:
To allow further computation of the actual HTTP call, the response object (`OAuth2::Response`) is stored on the `AccessToken`.

This solves #186.

I hope you are fine with this. I am happy to change the naming/implementation upon your suggestion. Thank you for your great work on this gem!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 All notable changes to this project will be documented in this file.
 
 ## [unreleased]
-- No significant changes.
+- Set the response object on the access token on Client#get_token (@cpetschnig)
 
 ## [1.3.0] - 2016-12-28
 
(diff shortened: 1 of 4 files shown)


cpetschnig:
CI is currently failing because of the changes introduced with the release of Rubocop 0.48 last night.

jimbali:
This was merged ages ago but it doesn't seem to have been released. It would be very useful.

pboling:
This was merged to master, so I think it is on the version 2.0 track, and there are still some outstanding issues that need to be addressed on master prior to releasing 2.0.  If you'd like to backport this PR to the [1-4-stable branch](https://github.com/oauth-xx/oauth2/tree/1-4-stable) then we can get it into a new release of 1.4.x, as I do not think this is a breaking change.

jimbali:
Thanks, I found a workaround for now. Looking forward to 2.0 though! 😊

PR #301, opened 2017-04-01
The gem should allow minor versions of Faraday releases

rhymes:
A gem should play nice with other gems in the Gemfile as much as possible.

Faraday is not breaking compatibility between minor releases, and there are many packages out there using faraday. Requiring faraday "less than the next minor release" means there's now way to upgrade Faraday in an app unless oauth2 releases a new gem each time which is a little extreme.

Thanks

### DIFF
--- a/oauth2.gemspec
+++ b/oauth2.gemspec
@@ -4,7 +4,7 @@ $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 require 'oauth2/version'
 
 Gem::Specification.new do |spec|
-  spec.add_dependency 'faraday', ['>= 0.8', '< 0.12']
+  spec.add_dependency 'faraday', ['>= 0.8', '~> 0.12']
   spec.add_dependency 'jwt', '~> 1.0'
   spec.add_dependency 'multi_json', '~> 1.3'
   spec.add_dependency 'multi_xml', '~> 0.5'


josephpage:
`~> 0.12` is quite redundant with `>= 0.8`
I would prefer `['>= 0.8', '< 1.0']` which prevents the gem from the next major release (more chance of breaking changes)

[A faraday contributor explains their policy as this](https://github.com/lostisland/faraday/issues/515#issuecomment-146046693) :
> When we're ready to release 1.0, then we'll know it is stable. Until then, the patch releases will signify bug fixes and minor releases will signify API upgrades, just like semver.

[Another comment from a maintainer](https://github.com/lostisland/faraday/issues/332#issuecomment-32550708) :
> Until 1.0.0 lands, you should lock to a specific patch level.

shaneog:
It would be great to get this change merged in!

dewetblomerus:
We would love to see this as we are blocked by this at work.

stanhu:
Can someone take a look at why Travis is failing here?

stanhu:
I think the `Gemfile` needs to be updated as well: https://github.com/intridea/oauth2/blob/master/Gemfile

shaneog:
[These lines](https://github.com/intridea/oauth2/blob/master/Gemfile#L3-L6) should ideally be replaced with just `gemspec` which will pull the dependencies defined in the `oauth2.gemspec`.

PR #302, opened 2017-04-03
Allow faraday 0.12.x

josephpage:
Faraday 0.12.0 was released 13 ago : https://github.com/lostisland/faraday/releases

⚠️ *Faraday is no more compatible with Ruby 1.8.7 since [0.10.0](https://github.com/lostisland/faraday/releases/tag/v0.10.0).*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 All notable changes to this project will be documented in this file.
 
 ## [unreleased]
-- No significant changes.
+- Add support for Faraday 0.12 (@rhymes, @josephpage)
 
 ## [1.3.0] - 2016-12-28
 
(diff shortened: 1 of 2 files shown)


rhymes:
I think this should be fixed "once and for all", see https://github.com/intridea/oauth2/pull/301

josephpage:
Yes... and no.
Yes because upgrades will become transparent.
No because Faraday doesn't respect Semver, so breaking changes can appear at any release.

rhymes:
@josephpage I understand, but the users of the oauth2 can still lock the version of faraday in their Gemfile if they need to. This way oauth2 does not need a release every time Faraday releases (which is quite often).

morgoth:
@sferik Can we unlock faraday 12.x please?

leematos:
What's going on with this one? Really blocked by not being able to upgrade this gem.

dewetblomerus:
We are getting blocked by this at work.

dblessing:
The PR at https://github.com/intridea/oauth2/pull/301 is better IMO. This one simply kicks the can down the road a month or two.

PR #299, opened 2017-03-28
Fix rubocop update changes

cpetschnig:
Since Rubocop release 0.48 the code style checks no longer pass on Travis

This PR was created basically by running `$ rubocop -a` and `$ rubocop --auto-gen-config`.

The rule for `Style/MixinGrouping` looks like it's buggy. There are already some PRs to fix it (https://github.com/bbatsov/rubocop/pull/4177, https://github.com/bbatsov/rubocop/pull/4176), but I thought it would be best to exclude it via the todo-file. Maybe this file can be removed once Rubocop 0.48.1 is released.

Also fixed some typos.

PR #291, opened 2017-02-07
Allow faraday 0.11 to be used

pschambacher:
Similar to https://github.com/intridea/oauth2/pull/277
Faraday 0.11 was released and ensures compatibility with Ruby 2.4 (which I added to travis.yml as well): https://github.com/lostisland/faraday/releases/tag/v0.11.0

Fixes #290

PR #303, opened 2017-04-15
Lock RuboCop version, fix configuration

bquorning:
PR #299 nicely fixes the style violations reported by the latest version of RuboCop. PR #292 did the same thing another time a new RuboCop release was causing red builds on Travis.

When contributing a feature or bugfix, you shouldn't have to deal with new style guide violations, or be tempted to merge when Travis is red (as in the case of #291). A better strategy (in my opinion) is to lock down the version of RuboCop, and update it in separate pull requests.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -60,6 +60,10 @@ Style/Lambda:
 Style/SpaceInsideHashLiteralBraces:
   EnforcedStyle: no_space
 
+# Disable for Ruby 1.9.3 compatibility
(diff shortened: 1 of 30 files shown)


josephpage:
#322 have done the same.
I think we should close this PR. @pboling

bquorning:
But the RuboCop version was unlocked again in 0e9093006dba577b9f73136d5cca5fc0c37dac22…

PR #305, opened 2017-04-30
Allow new access token class inheriting from AccessToken class to get access token instance of the its own class when refreshed

styd:
When a new class inherits from the AccessToken class, its refresh! method should return an instance of its own class, not of AccessToken's.


pboling:
@styd If you get a chance please rebase this on master.  We've improved many things in the codebase, and the CI build is now functional, and passing.

PR #308, opened 2017-05-29
Fix redirect_uri option : options need to be mutated

mathieulaporte:
Without this mutation the finale params look like : 
```ruby
{ ... 'redirect_uri' => 'http...', redirect_uri: 'http...' }
```
And the generated request has two time the param `redirect_uri`, probably not a bug for most servers but when you upgrade oauth2 all your webmocks fail...

For a better fix I think we should remove all mutations on the options attribute, but I don't imagine all side effects.

### DIFF
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -201,7 +201,7 @@ module OAuth2
     # @return [Hash] the params to add to a request or URL
     def redirection_params
       if options[:redirect_uri]
-        {'redirect_uri' => options[:redirect_uri]}
+        {'redirect_uri' => options.delete(:redirect_uri)}
       else
         {}
       end
--- a/spec/oauth2/client_spec.rb
+++ b/spec/oauth2/client_spec.rb
@@ -312,4 +312,19 @@ describe OAuth2::Client do
       expect(subject.connection.ssl.fetch(:ca_file)).to eq('foo.pem')
     end
   end
+
+  describe '#redirection_params' do
+    let(:uri) { 'http://test.com' }
+    let(:options) { {:redirect_uri => uri} }
+    subject { OAuth2::Client.new('abc', 'def', options) }
+    it 'mutates options' do
+      expect { subject.redirection_params }.to change { subject.options }
+    end
+
+    it 'remove the redirect_uri key in options' do
+      expect(subject.options).to include(:redirect_uri)
+      subject.redirection_params
+      expect(subject.options).to_not include(:redirect_uri)
+    end
+  end
 end


djberg96:
Instead of mutating the `options` hash, maybe just change 'redirect_uri' to a symbol? I generally try to avoid mutating external variables to avoid potential side effects.

pboling:
It would be super great if we could not mutate external variables.  PoLS, and all that. 👍

mastermatt:
I'm not sure why this change is needed. The only purpose of the `redirection_params` method is convert the `redirect_uri` key from a symbol to a string before sending it as extra query params to `Faraday::Connection#build_url`. But that method accepts either key type, and if both are provided it only uses one.

```ruby
2.4.2 :015 > conn = Faraday.new('http://github.com')
 => #<Faraday::Connection:0x00007fb2269a4bd0 @parallel_manager=nil, @headers={"User-Agent"=>"Faraday v0.12.2"}, @params={}, @options=#<Faraday::RequestOptions (empty)>, @ssl=#<Faraday::SSLOptions (empty)>, @default_parallel_manager=nil, @builder=#<Faraday::RackBuilder:0x00007fb2269a40e0 @handlers=[Faraday::Request::UrlEncoded, Faraday::Adapter::NetHttp]>, @url_prefix=#<URI::HTTP http://github.com/>, @proxy=nil>
2.4.2 :029 > conn.build_url('hello_world', {'redirect_uri' => 'http...uri_str.com', redirect_uri: 'http...uri_sym.com'}).to_s
 => "http://github.com/hello_world?redirect_uri=http...uri_sym.com"

2.4.2 :029 > conn.build_url('hello_world', {redirect_uri: 'http...uri_sym.com'}).to_s
 => "http://github.com/hello_world?redirect_uri=http...uri_sym.com"

2.4.2 :030 > conn.build_url('hello_world', {redirect_uri: 'http...uri_sym.com', 'redirect_uri' => 'http...uri_str.com'}).to_s
 => "http://github.com/hello_world?redirect_uri=http...uri_str.com"
```

It seems the right move would be just remove `redirection_params` altogether and pass the `params` as-is to `build_url`. The default query encoder in Faraday [converts all keys to strings here](https://github.com/lostisland/faraday/blob/aafbb24771d9127f8630072f897b1d325af6e828/lib/faraday/parameters.rb#L20). 
It is possible to provide a custom query encoder that makes a distinction between string and symbol keys, but I find it hard to think of a use case and this minor backwards compatibility break seems it could fit into the v2.0 nicely. 

I'll open a PR to continue the discussion of going the route of removing `redirection_params`.

pboling:
@mastermatt great analysis, but I am not understanding how the web mocks began failing with a doubled param if your analysis captures all leakage of the double param.

lib/oauth2/client.rb
```diff
@@ -201,7 +201,7 @@ def assertion
     # @return [Hash] the params to add to a request or URL
     def redirection_params
       if options[:redirect_uri]
-        {'redirect_uri' => options[:redirect_uri]}
+        {'redirect_uri' => options.delete(:redirect_uri)}
```
mathieujobin: how about....
```ruby
def redirection_params
  options.transform_keys!{ |key| key.to_s }
end
```

meganemura: Hash#transform_keys! is new feature in ruby 2.5.
We need to support [2.3 or higher in v2.x](https://github.com/oauth-xx/oauth2#rubies-with-continued-support-past-oauth2-2x).

pboling: @mathieujobin @meganemura We are not dropping any supported Rubies for v2.x.  We will continue to support down to Ruby 1.9 and JRuby 1.7.  We will drop Rubies < MRI 2.3 with version 3.0.0, which I expect to follow version 2.0.0 quickly.

mastermatt:
@pboling this PR should be closed to avoid further confusion.

pboling:
@mastermatt thanks!

pboling:
This is fixed in `master` branch as of #575

PR #310, opened 2017-06-07
Updated faraday dependency

nazarhussain:
To use latest and always latest faraday to avoid any http bugs that were fixed in faraday library.


nazarhussain:
All test seems to be failing on RubyCop syntax check. Is it normal, as I didn't make any change in to code.

PR #322, opened 2017-10-10
Lock Rubocop to 0.50

lautis:
Rubocop version updates introduce new rules that can break the build. It's not good to fail a build with unrelated changes because a new version of Rubocop was released. As a solution, lock Rubocop to the latest minor version and fix errors.

This caused build failure in https://github.com/intridea/oauth2/pull/318


lautis:
Didn't notice that there was already #303 that does the same. This PR manages with a bit less changes, though.

PR #325, opened 2017-10-23
Use `strict_encode64` for basic authentication

erik-escobedo:
`Base64.strict_encode64` method doesn't need to remove the `"\n"` characters a posteriori.

### DIFF
--- a/lib/oauth2/authenticator.rb
+++ b/lib/oauth2/authenticator.rb
@@ -31,7 +31,7 @@ module OAuth2
     end
 
     def self.encode_basic_auth(user, password)
-      'Basic ' + Base64.encode64(user + ':' + password).delete("\n")
+      'Basic ' + Base64.strict_encode64(user + ':' + password)
     end
 
   private


josephpage:
`Base64.strict_encode64` has been removed from code because ruby 1.8 doesn't have it.
https://github.com/oauth-xx/oauth2/commit/8799f123a9da1a824871178432e2763c58e89c5f

But the gem does not support this old ruby version anymore. 👍 

Could you apply this change to every calls of the method ?

pboling:
We will definitely not be supporting Ruby 1.8.  Need to think about support for other EOL'd Rubies.

erik-escobedo:
Totally. I’m on it!

josephpage:
FYI the rubocop offenses are fixed. So the Travis build should pass when you will push.

pboling:
@erik-escobedo ping!

pboling:
This change was merged in #414

PR #331, opened 2017-12-21
Change how connection is instantiated

spectator:
Currently, when running specs there is a warning `WARNING: Unexpected
middleware set after the adapter. This won't be supported from Faraday
1.0.`, if run with `OAUTH_DEBUG=true` then there is the same message
all over specs output, and in addition to that three errors like this

```ruby
  1) OAuth2::Client#request follows redirects properly
     Failure/Error: connection.response :logger, ::Logger.new($stdout) if ENV['OAUTH_DEBUG'] == 'true'

     Faraday::RackBuilder::StackLocked:
       can't modify middleware stack after making a request
     # ./lib/oauth2/client.rb:95:in `request'
     # ./lib/oauth2/client.rb:113:in `request'
     # ./spec/oauth2/client_spec.rb:205:in `block (3 levels) in <top (required)>'
```

This happens because logger for connection is enabled after afapter has
been already set when using `OAUTH_DEBUG=true`.

Fix is to instantiate connection object, set logger and then allow
changing that connection obejct with stuff like adapters.

(cherry picked from commit 432139f04fb9364b4720ce50f5dc3193ffe8e2d5)

### DIFF
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -53,14 +53,9 @@ module OAuth2
 
     # The Faraday connection object
     def connection
-      @connection ||= begin
-        conn = Faraday.new(site, options[:connection_opts])
-        if options[:connection_build]
-          conn.build do |b|
-            options[:connection_build].call(b)
-          end
-        end
-        conn
+      @connection ||= Faraday.new(site, options[:connection_opts]) do |builder|
+        builder.response(:logger, ::Logger.new($stdout)) if ENV['OAUTH_DEBUG'] == 'true'
+        options[:connection_build].call(builder) if options[:connection_build]
       end
     end
 
@@ -92,8 +87,6 @@ module OAuth2
     # @option opts [Symbol] :parse @see Response::initialize
     # @yield [req] The Faraday request
     def request(verb, url, opts = {}) # rubocop:disable CyclomaticComplexity, MethodLength, Metrics/AbcSize
-      connection.response :logger, ::Logger.new($stdout) if ENV['OAUTH_DEBUG'] == 'true'
-
       url = connection.build_url(url, opts[:params]).to_s
 
       response = connection.run_request(verb, url, opts[:body], opts[:headers]) do |req|
--- a/spec/oauth2/client_spec.rb
+++ b/spec/oauth2/client_spec.rb
@@ -44,10 +44,10 @@ describe OAuth2::Client do
     end
 
     it 'is able to pass a block to configure the connection' do
-      connection = double('connection')
       builder = double('builder')
-      allow(connection).to receive(:build).and_yield(builder)
-      allow(Faraday::Connection).to receive(:new).and_return(connection)
+
+      allow(Faraday).to receive(:new).and_yield(builder)
+      allow(builder).to receive(:response)
 
       expect(builder).to receive(:adapter).with(:test)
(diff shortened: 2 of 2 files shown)


jdelStrother:
@spectator what faraday version did you use this with?   I can't get this working at all with the latest faraday release (0.13.1).

Currently, something like this:
```ruby
client = OAuth2::Client.new("user", "pass", site: "http://example.com")
token = client.client_credentials.get_token({}, mode: :query, param_name: :oauth_token)
```
gives me the error:
```
/Users/jon/.gem/ruby/2.3.3/bundler/gems/oauth2-2ce3f5fb8048/lib/oauth2/response.rb:78:in `content_type': undefined method `values_at' for nil:NilClass (NoMethodError)
	from /Users/jon/.gem/ruby/2.3.3/bundler/gems/oauth2-2ce3f5fb8048/lib/oauth2/response.rb:84:in `parser'
	from /Users/jon/.gem/ruby/2.3.3/bundler/gems/oauth2-2ce3f5fb8048/lib/oauth2/response.rb:72:in `parsed'
	from /Users/jon/.gem/ruby/2.3.3/bundler/gems/oauth2-2ce3f5fb8048/lib/oauth2/error.rb:11:in `initialize'
	from /Users/jon/.gem/ruby/2.3.3/bundler/gems/oauth2-2ce3f5fb8048/lib/oauth2/client.rb:123:in `new'
	from /Users/jon/.gem/ruby/2.3.3/bundler/gems/oauth2-2ce3f5fb8048/lib/oauth2/client.rb:123:in `request'
	from /Users/jon/.gem/ruby/2.3.3/bundler/gems/oauth2-2ce3f5fb8048/lib/oauth2/client.rb:146:in `get_token'
	from /Users/jon/.gem/ruby/2.3.3/bundler/gems/oauth2-2ce3f5fb8048/lib/oauth2/strategy/client_credentials.rb:20:in `get_token'
	from foo.rb:20:in `<main>'
```

As far as I know, when passing a block to Faraday.new, you must specify middleware - Faraday won't default to its default middleware unless the block is missing.

How about something like https://github.com/jdelStrother/oauth2/commit/8419c7c248f699ac873963697c3b66addf1940f9 ?

spectator:
@jdelStrother I used the same Faraday version, but instantiating client with a block, I totally overlooked normal way and relied on tests. Your change looks good to me. Would you submit a pull request with your changes and a test?

/cc @pboling

jdelStrother:
The tests might be tricky - IIRC all the current ones pass a block so they can stub out the network request - but will see what I can do.

bughit:
Shouldn't this be backported and released? It's been in master for 4 years, meanwhile the latest release is still broken.

pboling:
Should is a complex word.

_None of the maintainers actively use the gem, unfortunately_.  We, (mostly the royal "Me"), are highly time constrained, and because of that, nearly all effort in maintenance is going toward prepping the v2.0 release.

*Please* work on backporting the fix if you need it.  The destination branch far backports into the v1.4 release train is available for anyone at `1-4-stable`.  That said, I may have some time to work on this in the next few weeks...

pboling:
Looks like the bug has now hit `1-4-stable`, and backporting this fix is now imperative.  https://github.com/oauth-xx/oauth2/issues/572.

Issue #332, opened 2018-01-03
Release 2.0.0 Checklist


pboling:
We have a new team of maintainers!

Myself (@pboling), @xtagon, @josephpage, @galois17 and a few others, are volunteering their efforts.

We are working on getting a new release out, @ *2.0.0*.  This is a checklist to make sure we do it in an orderly fashion.

- [x] What is the current minimum Ruby Version requirement of the gem, based on internal code and dependencies?
  - [x] [FIX] Documentation in `README`
  - [x] [FIX] Ruby req in `gemspec`
- [x] According to semver, how much of a version bump should the next release be?
  - [x] bumping faraday should make this at least a *minor* point release.
  - [x] officially dropping support for older Ruby versions might make this a *major* point release.

xtagon:
What's the process for determining the minimum supported Ruby version? Do we have visibility into what Ruby versions reverse dependencies are actually using?

josephpage:
From my research, here are the minimal ruby versions from the latest versions of the dependencies : 

- v1.9.3 Faraday - [Source (README)](https://github.com/lostisland/faraday#supported-ruby-versions)
- v2.1 JWT - [Source (gemspec)](https://github.com/jwt/ruby-jwt/blob/master/ruby-jwt.gemspec#L16)
- v1.8.7 multi_json [Source (travis.yml)](https://github.com/intridea/multi_json/blob/master/.travis.yml)
- v2.2.2 multi_xml [Source (README)](https://github.com/sferik/multi_xml/blob/master/README.md#supported-ruby-versions)
- v2.2 rack [Source (gemspec)](https://github.com/rack/rack/blob/master/rack.gemspec)

**multi_xml** should be optional, because not everybody wants to parse XML (and less APIs will be in XML in the future). **Rack** is only use for `Rack::Utils.parse_query` and `Rack::Utils.escape`. I think the `CGI` module in stdlib has alternative (`CGI::parse` and `CGI::escape`). For **JWT**, I just saw that the "Assertion" grant type is not mentioned in the final RFC.

Maybe we can hack something with the [Rubygems API](http://guides.rubygems.org/rubygems-org-api#gem-methods) to check the Gemfile of reverse dependencies.

pboling:
- [x] `Rack::Utils.escape`
[...]

PR #338, opened 2018-01-09
Replace Rack::Utils.escape by CGI.escape

josephpage:
Fixes the @pboling comment in #332.

> As of [May 2011](https://github.com/rack/rack/pull/140/files#diff-7c1a24d5b2fe58a6f925c7cacc6c55e7) and [continuing to today](https://github.com/jfirebaugh/rack/blob/0700cd04c41cb2e0882521fcf2fdca877b94e0ae/lib/rack/utils.rb#L14) Rack::Utils.escape calls CGI::escape directly, so switching would be fine.
> @pboling https://github.com/oauth-xx/oauth2/issues/332#issuecomment-356140908

### DIFF
--- a/lib/oauth2.rb
+++ b/lib/oauth2.rb
@@ -1,3 +1,7 @@
+# includes modules from stdlib
+require 'cgi'
+
+# includes gem files
 require 'oauth2/error'
 require 'oauth2/authenticator'
 require 'oauth2/client'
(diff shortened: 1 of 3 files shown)


pboling:
We should make sure we are loading 'cgi' from the standard Ruby library also, somewhere in the gem, like `require 'cgi'`...

pboling:
We are not currently loading it explicitly anywhere, and we should not rely on other libs (like Rack) to load it for us.
I believe this is where it should be loaded:
https://github.com/oauth-xx/oauth2/blob/master/lib/oauth2.rb

josephpage:
Done. Thanks @pboling

PR #342, opened 2018-01-11
Ruby 1.9.0 is the minimal supported version

josephpage:
Adds the minimal ruby version (1.9.0) in gemspec
Related to #332

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,6 @@
 AllCops:
   DisplayCopNames: true # Display the name of the failing cops
-  TargetRubyVersion: 2.1
+  TargetRubyVersion: 1.9
 
 Metrics/BlockLength:
   Enabled: false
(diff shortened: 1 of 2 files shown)


josephpage:
Yes :
```
Running RuboCop...
Inspecting 29 files
..C..........................
Offenses:
oauth2.gemspec:22:32: C: Gemspec/RequiredRubyVersion: required_ruby_version (1.9, declared in oauth2.gemspec) and TargetRubyVersion (2.1, declared in .rubocop.yml) should be equal.
  spec.required_ruby_version = '>= 1.9.0'
                               ^^^^^^^^^^
29 files inspected, 1 offense detected
RuboCop failed!
```

pboling:
Interesting... there a sensibility there.  With Rubocop targeting a post Ruby 2 version it will not fail on things like named parameters that may slip into the code.  Rather than turn off the check we should probably lower the target ruby version for Rubocop.

josephpage:
I've tried this alternative in my second commit.
But the current rubocop version doesn't support TargetRubyVersion under 2.1 : https://travis-ci.org/oauth-xx/oauth2/jobs/327620970#L1094

pboling:
Cool.  That *also* is understandable.  I'd hate to maintain a Ruby parser supporting so many versions of Ruby!  👍

PR #367, opened 2018-03-07
Stop rescuing parsing failure

### LINKED ISSUE #167: Hiding exceptions is dangerous
The plain rescue in MultiJson can potentially hide useful exceptions (intridea/multi_json#143)

``` ruby
:json  => lambda { |body| MultiJson.load(body) rescue body }, # rubocop:disable RescueModifier`
```

https://github.com/intridea/oauth2/blob/3d0ad2df0a3dd7e31b103e621dc3cfcd9895ca42/lib/oauth2/response.rb#L53

That's probably why rubocop complaints in the first place :frog:. 

Assuming the `rescue` is there for bad content-types and not to hide MultiJson errors, wouldn't it be better to ask the user to fix the issue himself by setting the parser?

pboling:
~There weren't even any specs testing that the rescue was happening, so it was never a *real* feature in the first place.~

Oddly, there are tests that are affected, just not in Ruby 2.5.0.  🗡 This will not be easy.

Fixes #167

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 
 ## [unreleased]
 
+- **Breaking**: No longer rescuing errors raised when parsing fails in MultiXml or MultiJson (@pboling)
 - Update testing infrastructure for all supported Rubies (@pboling and @josephpage)
 - **Breaking**: Set `:basic_auth` as default for `:auth_scheme` instead of `:request_body`. This was default behavior before 1.3.0. See [#285](oauth-xx/oauth2#285) (@tetsuya, @wy193777)
 - Token is expired if `expired_at` time is now (@davestevens)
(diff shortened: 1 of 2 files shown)


pboling:
I think this is an important idea, but the implementation is not straightforward.  The current code relies on this rescue, somehow, to make various Ruby interpreters behave the same way, and then it makes the ultimate final error raise optional via `:raise_errors`:
```
    # @option opts [Boolean] :raise_errors whether or not to raise an OAuth2::Error on 400+ status
    #   code response for this request.  Will default to client option
```
I am not sure how to remove it.  The build for this PR fails on some rubies, but not others when the rescue is removed.  And the way the tests fail is strange.  Some builds begin raising `MultiJson::ParseError`, but other continue to raise `OAuth2::Error`, and I don't understand how.

xtagon:
@pboling Hmmm, I'm not sure on this one. The second best thing to not rescuing would be to only rescue specific exceptions, but given the way the tests are behaving I'm not sure which ones would make sense.

pboling:
Closing this for now.  Not going to make it into the next release, need to redesign this throw catch behavior in a more fundamental way.

pboling:
This is now fixed in master branch, and will be in release `2.0.0`.  The spec suite is far more robust than ever before!

PR #376, opened 2018-04-11
Integration test and documentation for 2-legged Google use case

jhmoore:
As a followup to comments on https://github.com/oauth-xx/oauth2/pull/220, this is intended to provide some documentation and tests for using the `Assertion` strategy to retrieve an access token from Google via a service account.

Google's official documentation of this is here: https://developers.google.com/identity/protocols/OAuth2ServiceAccount

The tests, in addition to providing some illustration of the Google documentation as framed through the context of the gem, are also intended to provide some regression protection.  In particular, the previous naming collision of the URL parameter named `scope` with Google's requirement for a JWT claim key specifically named `scope` (which was most of the root cause of this gem's previous incompatibility with Google).

Also, some future paranoia on my part: I believe after reading extensively through too many RFCs that Google's implementation might actually conflict with [7523's requirement of a `sub` claim](https://tools.ietf.org/html/rfc7523#section-3).  Google uses the presence or lack of a `sub` claim to detect whether or not you are using your service account to impersonate or act on behalf of a particular user:

screen shot 2018-04-10 at 6 25 35 pm

I'm by no means an RFC-expert-reader-dude, but it would certainly seem to me that that is at odds with the RFC.

Anyway, right now this is fine, because in https://github.com/oauth-xx/oauth2/pull/220 I specifically _avoided_ validating the claimset for keys described as 'required' by that RFC.  I figured that this gem is sort of middleware, and it's up to providers to validate (or not!) their required claimset keys.  However . . . . if someone were to try to 'enhance' this gem in the future by having the bright idea to validate for 'required' keys . . . it would totally break the gem for anybody trying to use it to connect to Google via service accounts.  As such, this will hopefully provide a descriptive CI failure if somebody tries to go do that!


jhmoore:
@pboling Let me know what you think.  I am totally open to suggestions when it comes to spec directory structure -- I created `spec/examples` and `spec/fixtures` since I couldn't really find an existing place to put more integration-y tests like this.

Also happy to answer more questions or write more things illustrating the 2-legged stuff.

pboling:
@jhmoore This looks great.  Any idea about the JRuby spec failures?

jhmoore:
Interesting . . . yeah JRuby seems to have issues with opening that sample .p12 file for some reason.  I'll take a crack at figuring it out, but yeah it might be worth simply pending them if `OpenSSL` behaves differently somehow when using JRuby.

jhmoore:
@pboling well . . . I 'fixed the glitch,' so to speak.
```ruby
Sorry rescue 'Not sorry'
```
Haha I figure it'd be better to endure some janky setup and have the specs running in all environments than it would be to pend them out.  If anybody has JRuby and/or private key file expertise please educate me as to how to do it properly!

pboling:
I like it! 👍

PR #261, opened 2016-07-14
Avoid NoMethodError

ppworks:
If you use `faraday_middleware` and set `request` to `:json`,
you'll get **"NoMethodError: undefined method `encode' for #Hash:0x007ff31c8487a0"** in https://github.com/intridea/oauth2/blob/fb502c755ed8801aa5d4575f18b86a9b7ecf9f5e/lib/oauth2/error.rb#L28 .

This fixes avoid `NoMethodError` exception.

PR #380, opened 2018-05-02
Fix for OAuth2::Error trying to encode non-encodable objects

jhmoore:
We are pretty interested in seeing `2.0.0` released so we can unfork ourselves -- I was looking down the list of remaining items in https://github.com/oauth-xx/oauth2/milestone/1 and thought I might try to pitch in.

This is a replacement for https://github.com/oauth-xx/oauth2/pull/261 which looks dead.  The issue is trying to call `encode` on something which does not have an `encode` method.  However in trying to test (drive!) a fix for it, it seemed like there were no tests present at all for the `OAuth2::Error` class.  I created a new spec file and did my best to backfill specs for the already-existing functionality (of which there is quite a bit tacked on to the inherited `StandardError` behavior).

The GitHub diff of the error class itself looks kind of weird because of indentation, but a brief rundown of changes:
 - Made `#error_message` private.  Nothing outside this file talks to this method, and it sure seems like a "generate a formatted string for the error message" internal thing.
 - Renamed some locals to make me less nervous (e.g. `error_message` inside of `def error_message`, `message` inside a class descending from `StandardError`).
 - Made `#initialize` not rely on a local being defined only inside an `if` branch.  (Haha I know it works because Ruby is maaaagic, but y'know)
 - If `error_description` is present but not `error`, avoid displaying an errant `": "`.
 - If a formatted error description is not nil but is blank, avoid displaying a blank newline.
 - Backfilled specs for encoding and multi-line message generation.

Also: I was unable to determine why `@code` and `@description` are set and made into `attr_reader`s -- nothing outside the file seems to talk to those things either.  However, folks using this error in their implementations certainly could be relying on being able to call `error.code` or whatever.  I elected to leave them there (and _test_ that they are there), but open to suggestions on that front.

### DIFF
--- a/.rubocop_rspec.yml
+++ b/.rubocop_rspec.yml
@@ -21,3 +21,6 @@ RSpec/InstanceVariable:
 
 RSpec/NestedGroups:
   Enabled: false
+  
+RSpec/ExpectInHook:
+  Enabled: false
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ group :test do
     gem 'rubocop-rspec', '~> 1.24.0'
   end
   gem 'pry', '~> 0.11' if ruby_version >= Gem::Version.new('2.0')
+  gem 'rspec-pending_for'
 end
 
 # Specify non-special dependencies in oauth2.gemspec
--- a/gemfiles/jruby_1.7.gemfile
+++ b/gemfiles/jruby_1.7.gemfile
@@ -11,6 +11,7 @@ group :test do
   # gem "rubocop", "~> 0.41.2"
   gem 'rake'
   gem 'rspec'
+  gem 'rspec-pending_for'
 end
 
 gemspec :path => '../'
--- a/gemfiles/jruby_9.0.gemfile
+++ b/gemfiles/jruby_9.0.gemfile
@@ -3,6 +3,7 @@ source 'https://rubygems.org'
 group :test do
   gem 'rake'
   gem 'rspec'
+  gem 'rspec-pending_for'
   gem 'rubocop', '~> 0.53.0'
   gem 'rubocop-rspec', '~> 1.24.0'
 end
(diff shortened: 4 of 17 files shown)


jhmoore:
Oh now this is interesting.  The Travis builds in Ruby 1.9 and 2.0 are failing . . . . I believe that this behavior may have been broken in Ruby versions < 2.1 for some time but there was just no test to fail for it?

Apparently the behavior of `encode` [changed with 2.1](https://bibwild.wordpress.com/2013/03/12/removing-illegal-bytes-for-encoding-in-ruby-1-9-strings/).  Confirmed in consoles:

```ruby
2.0.0-p648 :001 > "\xAD bad!".encode("UTF-8", :invalid => :replace)
 => "\xAD bad!"
```

```ruby
2.3.3-p222 :001 > "\xAD bad!".encode("UTF-8", :invalid => :replace)
 => "� bad!" 
```

The JRuby builds are also failing in the same way as well.

pboling:
@jhmoore This is something I would vote for using [rspec-pending_for](https://github.com/pboling/rspec-pending_for) to skip those specs on Ruby < 2.1.

Then we can remove the caveat when we drop old Rubies in the following major release, 3.0.

jhmoore:
Hmm, looks like the JRuby versions are still running with MRI version numbers passed in.

Also haha 1.9.3 seems unable to even parse the file at all:
```
Failure/Error: load file
SyntaxError:
  /home/travis/build/oauth-xx/oauth2/spec/oauth2/error_spec.rb:76: invalid multibyte char (US-ASCII)
  /home/travis/build/oauth-xx/oauth2/spec/oauth2/error_spec.rb:76: invalid multibyte char (US-ASCII)
  /home/travis/build/oauth-xx/oauth2/spec/oauth2/error_spec.rb:76: syntax error, unexpected $end, expecting ')'
  ...ss subject.message.include?('??? invalid ???')
```

jhmoore:
@pboling looks like that did the trick!

pboling:
@jhmoore AFAIK, yes.  There may be other things lurking in issues, but I expect if we can clear those we'll be good to release.

jhmoore:
Ok cool, thanks.  I'll see what I can pick off there.

PR #390, opened 2018-06-07
Take time skew into consideration

klippx:
We have seen instances where `expired?` returns false but it actually is expired, and we think it might be due to time skew.

So, when checking if a token is expired, we should try to take time skew into consideration.

Ideally I would have had `issued_at` as part of the hash, since it should be read from the server. But for now I decided to calculate it (calculated_issued_at), I think it might be good enough.


klippx:
Closing this as the backwards calculation doesn't even work, we use Time.now +  expires_in - expires_in which gives Time.now again, which means clock skew is always zero.

Real fix in #391

PR #391, opened 2018-06-08
Take clock skew and min validity into consideration

klippx:
Real implementation of #390 (clock skew)

Changes
- Read issued at from the actual token instead of using Time.now (client time)
- Read expires_at from the actual token instead of using Time.now (client time) and adding expires_in to it
- Take clock skew into consideration when checking if token is expired

In case the token is not a JWT token with the expected keys, the old functionality is intact (use Time.now)

Thanks to @ashaeron for helping me :)


klippx:
Actually, I don't believe this is right. I need to think more about this...

Issue #393, opened 2018-06-14
Add built-in option for handling expiry latency


klippx:
We are using oauth2 in our Api.

When we make a request we use 
```ruby
  class Api
    def get(*args)
      with_error_handling(args) { connection.get(*args) }
    end
  end
```

The implementation of `connection` will construct the request headers, setting `'Authorization' => "Bearer #{access_token}"` where access_token is dynamically fetched:

```ruby
    # Manages the access_token (login, refresh)
    #
    # @return [String]
    def access_token
      if !token_store.get
        get_new_token
      elsif token_store.get.expired?
        begin
          refresh_token
        rescue OAuth2::Error
          get_new_token
        end
      end

      token_store.get.token
    end
```

`token_store` is a Concurrent::MutexAtomicReference instance in which we store the `OAuth2::AccessToken` instance, and retrieve using `get`. You can see we rely on `expired?` here.

The implementation of `with_error_handling` checks the response status, and if it is a 401 it will try again:

```ruby
    def with_error_handling(args = nil, retries = 1, &block)
      response = yield

      if response.status == 401 && retries.positive?
        return with_error_handling(args, retries - 1, &block)
      end
    end
```

The idea here is that we would generate a new header, invoke the call to `access_token` again, and this time the call to `expired?` would return true, and we would refresh the token. But because of latency, and how client sets expired to `Time.now + expires_in`, it seems to be not expired on client side but is actually expired on server side. So it will get a 401 again, since it is using the same bearer token.

We do not wish to force a token refresh in `with_error_handling` block in case of 401, we want the oauth2 client to recognize that the token is probably expired.

PR #394, opened 2018-06-14
Expire the token prematurely on client side

klippx:
Our assumption was that `expires_at?` is inaccurate and needs to be offset by some fixed latency amount that we never expect to exceed has been tested and this change solves the described issue #393 

screen shot 2018-06-14 at 10 30 17

The Kred::HttpErrors (which are due to getting two 401 in a row and failing the request) have vanished competely after the deploy (second green marker from the right).

The 10 seconds are a bit arbitrary, and in worst case (where latency is zero) will simply expire the token 10seconds too early on the client side. Which I think most people can live with since expires_at often is 300s or even 1800s.

### DIFF
--- a/lib/oauth2/access_token.rb
+++ b/lib/oauth2/access_token.rb
@@ -38,19 +38,21 @@ module OAuth2
     # @option opts [String] :header_format ('Bearer %s') the string format to use for the Authorization header
     # @option opts [String] :param_name ('access_token') the parameter name to use for transmission of the
     #    Access Token value in :body or :query transmission mode
-    def initialize(client, token, opts = {}) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+    def initialize(client, token, opts = {}) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, Metrics/CyclomaticComplexity
       @client = client
       @token = token.to_s
       opts = opts.dup
       [:refresh_token, :expires_in, :expires_at, :expires_latency].each do |arg|
         instance_variable_set("@#{arg}", opts.delete(arg) || opts.delete(arg.to_s))
       end
-      @expires_latency ||= 0
       @expires_in ||= opts.delete('expires')
       @expires_in &&= @expires_in.to_i
       @expires_at &&= @expires_at.to_i
       @expires_latency &&= @expires_latency.to_i
-      @expires_at ||= Time.now.to_i + @expires_in - @expires_latency if @expires_in
+      if @expires_in
+        @expires_at ||= Time.now.to_i + @expires_in
+        @expires_at -= @expires_latency if @expires_latency
+      end
       @options = {:mode          => opts.delete(:mode) || :header,
                   :header_format => opts.delete(:header_format) || 'Bearer %s',
                   :param_name    => opts.delete(:param_name) || 'access_token'}
--- a/lib/oauth2/access_token.rb
+++ b/lib/oauth2/access_token.rb
@@ -1,5 +1,6 @@
 module OAuth2
   class AccessToken
+    REDUCE_VALIDITY = 10
     attr_reader :client, :token, :expires_in, :expires_at, :params
     attr_accessor :options, :refresh_token, :response
 
@@ -69,10 +70,12 @@ module OAuth2
     end
 
     # Whether or not the token is expired
+    # - The client will expire the token REDUCE_VALIDITY seconds
+    #   prematurely to offset latency issues
     #
     # @return [Boolean]
     def expired?
-      expires? && (expires_at <= Time.now.to_i)
+      expires? && (expires_at - REDUCE_VALIDITY <= Time.now.to_i)
     end
 
     # Refreshes the current Access Token
--- a/lib/oauth2/access_token.rb
+++ b/lib/oauth2/access_token.rb
@@ -38,7 +38,7 @@ module OAuth2
     # @option opts [String] :header_format ('Bearer %s') the string format to use for the Authorization header
     # @option opts [String] :param_name ('access_token') the parameter name to use for transmission of the
     #    Access Token value in :body or :query transmission mode
-    def initialize(client, token, opts = {}) # rubocop:disable Metrics/AbcSize
+    def initialize(client, token, opts = {}) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
       @client = client
       @token = token.to_s
       opts = opts.dup
--- a/lib/oauth2/access_token.rb
+++ b/lib/oauth2/access_token.rb
@@ -1,7 +1,6 @@
 module OAuth2
   class AccessToken
-    REDUCE_VALIDITY = 10
-    attr_reader :client, :token, :expires_in, :expires_at, :params
+    attr_reader :client, :token, :expires_in, :expires_at, :expires_latency, :params
     attr_accessor :options, :refresh_token, :response
 
     class << self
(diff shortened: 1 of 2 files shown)


pboling:
Interesting solution.

What if instead of adding a constant that would globally apply for every use case, including some where it may be detrimental, we added the `expiry_reduction` as an option to `AccessToken#initialize`:

Current:
```
      opts = opts.dup
      [:refresh_token, :expires_in, :expires_at].each do |arg|
        instance_variable_set("@#{arg}", opts.delete(arg) || opts.delete(arg.to_s))
      end
      @expires_in ||= opts.delete('expires')
      @expires_in &&= @expires_in.to_i
      @expires_at &&= @expires_at.to_i
      @expires_at ||= Time.now.to_i + @expires_in if @expires_in
```

Update:
```
      opts = opts.dup
      [:refresh_token, :expires_in, :expires_at, :expiry_reduce].each do |arg|
        instance_variable_set("@#{arg}", opts.delete(arg) || opts.delete(arg.to_s))
      end
      @expires_in ||= opts.delete('expires')
      @expires_in &&= @expires_in.to_i
      @expires_at &&= @expires_at.to_i
      @expiry_reduce &&= @expiry_reduce.to_i
      @expires_at ||= Time.now.to_i + @expires_in if @expires_in
      @expires_at -= @expiry_reduce if @expiry_reduce
```

Then the handling of the expires_at can remain the same, from the perspective of all outside callers of this class, the expiry is just sooner than it otherwise would have been.

Would that work for your use case?

klippx:
@pboling Sounds good! I updated with inspiration from your pseudo code, and added some specs.

lib/oauth2/access_token.rb
```diff
@@ -32,22 +32,25 @@ def from_kvform(client, kvform)
...
+      [:refresh_token, :expires_in, :expires_at, :expires_latency].each do |arg|
         instance_variable_set("@#{arg}", opts.delete(arg) || opts.delete(arg.to_s))
       end
+      @expires_latency ||= 0
       @expires_in ||= opts.delete('expires')
       @expires_in &&= @expires_in.to_i
       @expires_at &&= @expires_at.to_i
-      @expires_at ||= Time.now.to_i + @expires_in if @expires_in
+      @expires_latency &&= @expires_latency.to_i
+      @expires_at ||= Time.now.to_i + @expires_in - @expires_latency if @expires_in
```
pboling: This is almost correct.
The instance variables initially get set to a value if provided, or `nil`.
```
instance_variable_set("@#{arg}", opts.delete(arg) || opts.delete(arg.to_s))
```
The `&&=` then ensures that the values *stay* `nil`, or if they are String/Integer, they are cast to integer.
```
@expires_in &&= @expires_in.to_i
@expires_at &&= @expires_at.to_i
@expires_latency &&= @expires_latency.to_i
```
This means that any of the values can be `nil`, and if nil, can't be used in math, unless cast to integer, or protected by a condition.
```
@expires_at ||= Time.now.to_i + @expires_in if @expires_in
@expires_at -= @expires_latency if @expires_latency
```
The other difference between this and what you have, is for this version the `@expires_latency` is always applied, if set, even if an exact `:expires_at` time was provided.  I think that this is in  line with the underlying purpose of `@expires_latency` though.

lib/oauth2/access_token.rb
```diff
@@ -32,22 +32,25 @@ def from_kvform(client, kvform)
...
-    def initialize(client, token, opts = {}) # rubocop:disable Metrics/AbcSize
+    def initialize(client, token, opts = {}) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
       @client = client
       @token = token.to_s
       opts = opts.dup
-      [:refresh_token, :expires_in, :expires_at].each do |arg|
+      [:refresh_token, :expires_in, :expires_at, :expires_latency].each do |arg|
         instance_variable_set("@#{arg}", opts.delete(arg) || opts.delete(arg.to_s))
       end
+      @expires_latency ||= 0
```
pboling: Oh, I see you are setting it to `0` by default.  I think I'd prefer having it be `nil` by default in keeping with the other options' `nil` defaults.

spec/oauth2/access_token_spec.rb
```diff
@@ -75,6 +75,27 @@ def assert_initialized_token(target) # rubocop:disable Metrics/AbcSize
...
+        hash = {:access_token => token, :expires_latency => '10'}
+        target = described_class.from_hash(client, hash)
+        expect(target.expires_latency).to eq 10
+      end
+
+      it 'sets it 0 by default' do
+        hash = {:access_token => token, :expires_in => 100}
+        target = described_class.from_hash(client, hash)
+        expect(target.expires_latency).to eq 0
+      end
```
pboling: Would prefer that it be `nil` by default.

spec/oauth2/access_token_spec.rb
```diff
@@ -75,6 +75,27 @@ def assert_initialized_token(target) # rubocop:disable Metrics/AbcSize
...
+
+      it 'reduces expires_at by the given amount' do
+        allow(Time).to receive(:now).and_return(1_530_000_000)
+        hash = {:access_token => token, :expires_latency => 10, :expires_in => 100}
+        target = described_class.from_hash(client, hash)
+        expect(target.expires_at).to eq 1_530_000_090
+      end
```
pboling: The code as you have it would fail a spec where `expires_at` and `expires_latency` are both provided explicitly, as `expires_at` would not be adjusted by the `expires_latency`.  See my comment inline with the code for a suggestion at a small change to make `expires_latency` globally applicable when provided.

pboling:
I like the direction of this!

spec/oauth2/access_token_spec.rb
klippx: - [x] Added a spec for this scenario

spec/oauth2/access_token_spec.rb
klippx: - [x] make it nil by default

klippx:
Made the initialize code slightly more complex and had to add another rubocop exclusion I'm afraid. The initialize code is getting hairy, I have to admit.

lib/oauth2/access_token.rb
```diff
@@ -32,22 +32,27 @@ def from_kvform(client, kvform)
...
         instance_variable_set("@#{arg}", opts.delete(arg) || opts.delete(arg.to_s))
       end
       @expires_in ||= opts.delete('expires')
       @expires_in &&= @expires_in.to_i
       @expires_at &&= @expires_at.to_i
-      @expires_at ||= Time.now.to_i + @expires_in if @expires_in
+      @expires_latency &&= @expires_latency.to_i
+      if @expires_in
+        @expires_at ||= Time.now.to_i + @expires_in
+        @expires_at -= @expires_latency if @expires_latency
```
pboling: I would expect this setting to always apply if provided.  I don't see a reason to have it be dependent on the `@expires_in` being set.

```
@expires_at ||= Time.now.to_i + @expires_in if @expires_in
@expires_at -= @expires_latency if @expires_latency
```

Was there a reason for structuring it this way?

I prefer to have features that work independent of other features, for less confusing, less complex logic.

spec/oauth2/access_token_spec.rb
```diff
@@ -75,6 +75,41 @@ def assert_initialized_token(target) # rubocop:disable Metrics/AbcSize
...
+        expect(target.expires_latency).to be_nil
+      end
+
+      it 'reduces expires_at by the given amount' do
+        allow(Time).to receive(:now).and_return(expires_at)
+        target = described_class.from_hash(client, hash)
+        expect(target.expires_at).to eq(expires_at + expires_in - expires_latency)
+      end
+
+      it 'reduces expires_at by the given amount if expires_at is provided as option' do
```
pboling: Perhaps one more additional test - `reduces expires_at by the given amount if expires_at is not provided as option`?

klippx: I think `reduces expires_at by the given amount` covers this case since `expires_at` is not provided by default and this test explicitly says that it is provided, and adds it to the hash.

klippx:
I added further changes as per review. Unfortunately the rubocop complaints are still there, I cannot get rid of them without a bigger refactoring of the code. Perhaps this can be raised as separate issue.

klippx:
Need to keep this branch open a while until a new version of the gem is released

pboling:
Progress is slow on the next release, but it is coming!

PR #399, opened 2018-07-11
Fix duplicate redirect_uri in get_token request.

markus:
OAuth2::Client#redirection_params converts the symbol key :redirect_uri
to a string, so we have to do the same for params in get_token before merging
to avoid submitting both :redirect_uri and "redirect_uri".

Noticed this issue when using the linkedin-oauth2 gem.

### DIFF
--- a/lib/oauth2/strategy/auth_code.rb
+++ b/lib/oauth2/strategy/auth_code.rb
@@ -26,6 +26,9 @@ module OAuth2
       # @note that you must also provide a :redirect_uri with most OAuth 2.0 providers
       def get_token(code, params = {}, opts = {})
         params = {'grant_type' => 'authorization_code', 'code' => code}.merge(@client.redirection_params).merge(params)
+        params.keys.each do |key|
+          params[key.to_s] = params.delete(key) if key.is_a?(Symbol)
+        end
 
         @client.get_token(params, opts)
       end
--- a/spec/oauth2/strategy/auth_code_spec.rb
+++ b/spec/oauth2/strategy/auth_code_spec.rb
@@ -7,6 +7,7 @@ RSpec.describe OAuth2::Strategy::AuthCode do
   let(:kvform_token) { 'expires_in=600&access_token=salmon&refresh_token=trout&extra_param=steve' }
   let(:facebook_token) { kvform_token.gsub('_in', '') }
   let(:json_token) { MultiJson.encode(:expires_in => 600, :access_token => 'salmon', :refresh_token => 'trout', :extra_param => 'steve') }
+  let(:redirect_uri) { 'http://example.com/redirect_uri' }
 
   let(:client) do
     OAuth2::Client.new('abc', 'def', :site => 'http://api.example.com') do |builder|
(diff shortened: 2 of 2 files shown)


markus:
@pboling thanks for your review. I fixed the pull request to convert all symbol keys to strings on the copy of params.

swrobel:
I don't understand why, but this change is breaking [omniauth-stripe-connect](https://github.com/isaacsanders/omniauth-stripe-connect). If I use the following in my Gemfile (previous commit, upgrade Farday):

```ruby
gem 'oauth2', github: 'oauth-xx/oauth2', ref: '5972a99'
```

Then the OAuth process completes fine. If I update my Gemfile to include this commit:

```ruby
gem 'oauth2', github: 'oauth-xx/oauth2', ref: '48cd561'
```

Then I get the following in my logs and OAuth fails:

```
I, [2018-09-24T11:37:25.793220 #65433]  INFO -- omniauth: (stripe_connect) Request phase initiated.
I, [2018-09-24T11:37:57.866313 #65433]  INFO -- omniauth: (stripe_connect) Callback phase initiated.
E, [2018-09-24T11:37:58.120554 #65433] ERROR -- omniauth: (stripe_connect) Authentication failure! invalid_credentials: OAuth2::Error, invalid_client: No such API key: ca_XXX
{
  "error": "invalid_client",
  "error_description": "No such API key: ca_XXX"
}
```

swrobel:
Possibly related to [this line](https://github.com/isaacsanders/omniauth-stripe-connect/blob/402624719a41a79f9af41822fdabe2452557ffde/lib/omniauth/strategies/stripe_connect.rb#L55) in omniauth-stripe-connect?

pboling:
I think you nailed it.  As it stands now current master breaks compatibility with some use cases involving symbols.
Seems like we have a few options:
1. Add a warning when a symbol key is detected, as it can result in unexpected behavior.
2. Use a [MultiKeyHash that has both keys](https://github.com/SamirHodzic/multi_key_hash), symbol and string, pointing to the same value.
@swrobel what are your thoughts?

markus:
In my opinion it would be nice if warnings would not be needed.

pboling:
A multi_key_hash won't work, because the keys are not pointing to the same value.
```
>> require 'multi_key_hash'
=> true
>> a = MultiKeyHash.new([:a, 'a'] => 1)
=> #<MultiKeyHash:0x007fe100874720 @next_key=2, @inner={:a=>"7913b04c5", "7913b04c5"=>1, "a"=>"7913b04c5"}, @outer={:a=>"7913b04c5", "7913b04c5"=>1, "a"=>"7913b04c5"}>
>> a[:a]
=> 1
>> a['a']
=> 1
>> a['a'] = 2
=> 2
>> a[:a]
=> 1
```

So it would have to be a solution like ActiveSupport's `HashWithIndifferentAccess`.

pboling:
This is the gem we would need to use:
https://github.com/TamerShlash/symbolized
```
>> require 'symbolized'
=> true
>> h = { 'a' => 'b' }
=> {"a"=>"b"}
>> h['a']
=> "b"
>> h[:a]
=> nil
>> h.keys #=> ['a']
=> ["a"]
>> symbolized_hash = h.to_symbolized_hash
=> {:a=>"b"}
>> symbolized_hash['a']
=> "b"
>> symbolized_hash[:a]
=> "b"
>> symbolized_hash.keys
=> [:a]
>> symbolized_hash.key?('a')
=> true
>> symbolized_hash.key?(:a)
=> true
```

PR #408, opened 2018-09-11
Fixed expires_at for formatted time

Lomey:


pboling:
The documentation for the argument is:
```
     # @option opts [FixNum, String] :expires_at (nil) the epoch time in seconds in which AccessToken will expire
```
Is there a reason the client code you have implemented can't transform the value prior to passing it in?

Lomey:
There is not possibility to convert value prior to passing it in, because request and parsing is going in gem code.
my code:
 `data = client.auth_code.get_token(params[:code], token_params || {})`
where client is instance of gem class. And other is gem methods
```
def client
      @client ||= OAuth2::Client.new(client_id, secret_key,
                                     site: url, authorize_url: authorize_url, token_url: token_url)
    end
```
I will write additional spec, ok?

meganemura:
Found typo in pull request title.

greasysock:
Is there anything else that needs to be done before this gets merge this in? I am attempting to use square and am having the same issue.

pboling:
@Lomey @greasysock @meganemura I have released 1.4.4 with this change.  It is also now on master to be included with future 2.0 release.

dobon:
This PR has introduced a **breaking regression error**:
Consider the case when expired_at is a timestamp (string) and IS expired. EG: `{expired_at: "1577836800"}` (timestamp = 2020-01-01 00:00:00 +0000)

Old behavior: `@expires_at` is assigned the **string** value `'1577836800'` at `lib/oauth2/access_token.rb ln:46`, then cast as an integer on `ln:50`. The value of `@expires_at` is the **integer** timestamp `1577836800`.

New breaking behavior: `@expires_at` is assigned the **string** value `'1577836800'` at `lib/oauth2/access_token.rb ln:46`, then is sent to the `convert_expires_at` method on `ln:50`. The `convert_expires_at` method creates a temporary variable `expires_at_i` which is assigned the integer-cast value of `@expires_at` (ie: `expires_at_i == '1577836800'.to_i == 1577836800`). This supposed timestamp is compared against the current timestamp and because the token is expired, execution continues to ln:182. `@expires_at` is a string, so this code assumes that it is a date string, but it is actually a timestamp string, so parsing time fails with `ArgumentError (argument out of range)`

**The `convert_expires_at` method introduces BREAKING changes and should be reverted or amended.**

Possible solution: revert this code and fix the Stripe gem instead. This makes more sense, rather than baking special cases into the library, adjust the consumer to conform their usage to the library.

Possible solution: do better check than "is `expires_at` a string?". This could mean a regex to check date string format.

An expired timestamp string for `@expires_at` should be added to the spec. Actually, this PR edited the spec to remove the case where this happens, replacing an expired timestamp string with a future timestamp string, which circumvents the problem, while also demonstrating that the developer must have been aware of the issue. =/

Lomey:
@dobon, I am going to fix this asap

dobon:
This should work (using iso8601 specifically, rather than generic parse, but it could use generic parse as well):
```
 def convert_expires_at(expires_at)
    begin
       Time.iso8601(expires_at).to_i
    rescue ArgumentError
       expires_at.to_i
    end      
end
```

And then add specs for ISO8601 future + expired, integer future + expired, timestamp (ie: integer string) future + expired.

Lomey:
@dobon Have opened the pull request https://github.com/oauth-xx/oauth2/pull/532

dobon:
Thanks @Lomey much appreciated. This looks great. I've left some feedback on the new PR:
1. add spec for integer expires_at (rather than just iso8601 string and timestamp strings): provided missing test in github comment on PR
2. identify gap in timestamp string regex: recommended the revision of `convert_expires_at` I wrote above instead of the version on the PR.

I would make a PR myself, but this project has some rules about contributor, so I'll just pass along my feedback to you. I hope you don't mind. Really appreciate your effort on this.

pboling:
@dobon Would love to have you as a contributor, if you are interested.  We need more hands and minds working on this gem! ❤️

dobon:
@pboling I've written my own PR to get this fixed: https://github.com/oauth-xx/oauth2/pull/533

If approved, then https://github.com/oauth-xx/oauth2/pull/532 can be closed.

pboling:
@dobon This change, which introduced a bug, was the main feature of the 1.4.4 release of the gem, which is the current version of the gem.  We need your fix in #533 to be back ported to `1-4-stable` branch to fix the regression.

PR #410, opened 2018-09-20
Fix for circular reference exceptions between Error and Response classes

jhmoore:
This is to replace https://github.com/oauth-xx/oauth2/pull/201.  This is a breaking change!  It removes the ability to call `.error` from an `OAuth2::Response` object.

Also, adding some test coverage and fixing a potential 500 if a response is encountered with no headers.  The `Response` class was calling `response.headers.values_at` without checking to see if `response.headers` returned a valid thing, and also there were no tests for `#content_type`.

As an aside, the exception referenced in that older pull request only is a thing in [older versions of Rails](https://apidock.com/rails/ActiveSupport/JSON/Encoding/CircularReferenceError).  The current version of ActiveSupport JSON encoding says:

```
"The JSON encoder in Rails 4.1 no longer offers protection from circular references. " \
"You are seeing this warning because you are rescuing from (or otherwise referencing) " \
"ActiveSupport::Encoding::CircularReferenceError. In the future, this error will be " \
"removed from Rails. You should remove these rescue blocks from your code and ensure " \
"that your data structures are free of circular references so they can be properly " \
"serialized into JSON.\n\n" \
"For example, the following Hash contains a circular reference to itself:\n" \
"   h = {}\n" \
"   h['circular'] = h\n" \
"In this case, calling h.to_json would not work properly."
```

It is likely not ideal to raise stack level too deep levels _either_, but from a practical testing standpoint it is not a good idea to detect the presence of that old error.

### DIFF
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -115,7 +115,6 @@ module OAuth2
       when 400..599
         error = Error.new(response)
         raise(error) if opts.fetch(:raise_errors, options[:raise_errors])
-        response.error = error
         response
       else
         error = Error.new(response)
--- a/lib/oauth2/error.rb
+++ b/lib/oauth2/error.rb
@@ -5,7 +5,6 @@ module OAuth2
     # standard error codes include:
     # 'invalid_request', 'invalid_client', 'invalid_token', 'invalid_grant', 'unsupported_grant_type', 'invalid_scope'
     def initialize(response)
-      response.error = self
       @response = response
       message_opts = {}
 
--- a/lib/oauth2/response.rb
+++ b/lib/oauth2/response.rb
@@ -6,7 +6,7 @@ module OAuth2
   # OAuth2::Response class
   class Response
     attr_reader :response
-    attr_accessor :error, :options
+    attr_accessor :options
 
     # Procs that, when called, will parse a response body according
     # to the specified format.
--- a/spec/oauth2/client_spec.rb
+++ b/spec/oauth2/client_spec.rb
@@ -284,7 +284,6 @@ RSpec.describe OAuth2::Client do
 
       expect(response.status).to eq(401)
       expect(response.headers).to eq('Content-Type' => 'application/json')
-      expect(response.error).not_to be_nil
     end
 
     %w[/unauthorized /conflict /error /different_encoding /ascii_8bit_encoding].each do |error_path|
@@ -315,7 +314,7 @@ RSpec.describe OAuth2::Client do
       begin
         subject.request(:get, '/error')
       rescue StandardError => e
-        expect(e.response).not_to be_nil
+        expect(e.response).to be_a(OAuth2::Response)
         expect(e.to_s).to match(/unknown error/)
       end
     end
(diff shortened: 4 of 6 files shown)


jhmoore:
Hah.  Well, it seems like the reason I couldn't make any tests fail is that the pattern in the `OAuth2::Response` specs of making a `double` for the response object behaves differently.  When attached to the `OAuth2::Error` (and vice versa), the double does _not_ get to-jsoned in the same way that a 'real' response object (i.e. `Faraday::Response`) would.

Calling `to_json` on an `OAuth2::Error` with a test double response has no error.  However, with a real one:

```
2.3.3 :001 > raw_response = Faraday::Response.new(:body => "hello", :response_headers => {"stuff" => "things"})
 => #<Faraday::Response:0x007fb619052ec8 @on_complete_callbacks=[], @env=#<Faraday::Env @body="hello" @response_headers={"stuff"=>"things"}>>
2.3.3 :002 > response_object = OAuth2::Response.new(raw_response)
 => #<OAuth2::Response:0x007fb619048338 @response=#<Faraday::Response:0x007fb619052ec8 @on_complete_callbacks=[], @env=#<Faraday::Env @body="hello" @response_headers={"stuff"=>"things"}>>, @options={:parse=>:automatic}> 
2.3.3 :003 > error = OAuth2::Error.new(response_object)
 => #<OAuth2::Error: hello> 
2.3.3 :004 > error.to_json
SystemStackError: stack level too deep
```

pboling:
The next release, version 2.0, is not intended to drop support for any existing supported versions of Rails, or Ruby.  The version support dropping release is milestone 3.0.  So we just need to be careful that we don't break support for the currently supported versions.  *Breaking changes* are acceptable though, it is a major SemVer version bump.

jhmoore:
@pboling agreed -- I think testing against specifically trying to prevent an old error is not the correct approach, but we also want to allow for it.  

I'm just a little baffled that I can't seem to reproduce any error in test.  I reproduced a stack level too deep error using the gem in the context of our app (calling `to_json` on the response object), but doing the same in the gem's rspec suite does not seem to result in an error.  I am admittedly far from an expert on Ruby's `to_json` meta-magic, but it seems strange.

jhmoore:
Ok after talking around and around about this in the office, the consensus is:

 - Given the choice of `error.response` or `response.error`: it is probably more useful to be able to access `.response` from an instance of a raised exception that you've rescued from.  We ourselves do this a couple of times in our app.  While it could be worked around (i.e. look at the message of the error), it is nice to be able to just get the integer of the status code.
 - It seems _less_ weird to set `@response` on `OAuth2::Error` than it is for `OAuth2::Error` to set itself as `.error` on the response argument object and mutate it.
 - There actually were a couple of behaviors tested around accessing the response object from the error, but not so much the other way around.

Totally open to discussion though, if anyone has any better ideas!

pboling:
I am fully willing to accept your reasoning.  It seems perfectly *reasonable*.

jhmoore:
Hahaha sounds good.  I am still a little weirded out that I couldn't reproduce the circular reference error in test, but in theory there are now specs at least documenting it.  I think this is ready to go.

PR #414, opened 2018-10-04
Use Base64.strict_encode64

meganemura:
This is a successor of #325.

### DIFF
--- a/lib/oauth2/authenticator.rb
+++ b/lib/oauth2/authenticator.rb
@@ -31,7 +31,7 @@ module OAuth2
     end
 
     def self.encode_basic_auth(user, password)
-      'Basic ' + Base64.encode64(user + ':' + password).delete("\n")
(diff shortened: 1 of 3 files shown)


meganemura:
@josephpage @pboling Would you please review this?

pboling:
I need to review this for breaking changes.

We are not dropping any existing support for old Rubies in the 2.0.0 release.

```
# Base64.strict_encode64 is not available on Ruby 1.8.7
```

This might need to go into the follow up 3.0.0 release.

meganemura:
@pboling 

In https://github.com/oauth-xx/oauth2/milestone/1 and #325, 

> 2.x is the last version series that will have Ruby 1.9/2.0/2.1/2.2 support.

So, I think 2.x release does not need to support ruby 1.8.

pboling:
Oh good point!  I need to add a Ruby support matrix to the Readme.

PR #426, opened 2018-10-14
Fix a link in changelog

meganemura:
title only.


CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 ## [unreleased]
```
willnet: @meganemura This line should be also fixed because v1.4.1 was released

meganemura: @willnet It will be updated by #424.

willnet: 👍

PR #433, opened 2018-11-15
Fix OAuth2::Client#request send an incorrect query.

asm256:
If field name have square brackets and numbers in params options,
request will send an incorrect query.

### DIFF
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -93,8 +93,9 @@ module OAuth2
     # @option opts [Symbol] :parse @see Response::initialize
     # @yield [req] The Faraday request
     def request(verb, url, opts = {}) # rubocop:disable CyclomaticComplexity, MethodLength, Metrics/AbcSize
-      url = connection.build_url(url, opts[:params]).to_s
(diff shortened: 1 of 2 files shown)


lib/oauth2/client.rb
```diff
@@ -93,8 +93,9 @@ def token_url(params = nil)
     # @option opts [Symbol] :parse @see Response::initialize
     # @yield [req] The Faraday request
     def request(verb, url, opts = {}) # rubocop:disable CyclomaticComplexity, MethodLength, Metrics/AbcSize
-      url = connection.build_url(url, opts[:params]).to_s
+      url = connection.build_url(url).to_s
       response = connection.run_request(verb, url, opts[:body], opts[:headers]) do |req|
+        req.params.update(opts[:params]) if opts[:params]
```
pboling: I am not sure if this is the fix we want to apply, but the specs are great!  Thanks for creating them.  It will help us identify the cause of the issue, and think about how to fix it in the best way.

pboling: This is the fix. My fears were unfounded!

PR #442, opened 2019-01-26
Allow users of this gem to override the logger

rthbound:
- This gem supports an existing OAUTH_DEBUG feature where log messages
  would be printed to $stdout when an ENV['OAUTH_DEBUG'] environment
  flag is raised.
  - Previously, the logging (if enabled) was hardcoded so as to always write to $stdout.
  - With this change, the gem will offer users the ability to override
  the logger used. This should provide much more freedom, allowing users
  to more easily wrangle debug output by providing other loggers and
  using other log formatters.

### DIFF
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -24,6 +24,7 @@ module OAuth2
     # @option opts [Hash] :connection_opts ({}) Hash of connection options to pass to initialize Faraday with
     # @option opts [FixNum] :max_redirects (5) maximum number of redirects to follow
     # @option opts [Boolean] :raise_errors (true) whether or not to raise an OAuth2::Error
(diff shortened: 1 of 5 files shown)


lib/oauth2/client.rb
```diff
@@ -39,7 +40,8 @@ def initialize(client_id, client_secret, options = {}, &block)
                   :connection_opts  => {},
                   :connection_build => block,
                   :max_redirects    => 5,
-                  :raise_errors     => true}.merge(opts)
+                  :raise_errors     => true,
+                  :logger           => ::Logger.new($stdout)}.merge!(opts)
```
rthbound: There is one small, unrelated change happening here that may be difficult to spot:

```
# before
opts = {}.merge(opts)

# after, #merge! modifies original hash rather than instantiating another
opts = {}.merge!(opts)
```

rthbound:
@pboling the first commit in this PR should repair the build

Issue #459, opened 2019-07-03
Client `fetch_token` throwing `NoMethodError`


joaolrpaulo:
I am currently using this library to fetch tokens from a centralized identifier service.

The problem occurs when the identifier service is unreachable and we are passing the `:raise_errors` option set to `false`.

The `fetch_token` method throws `NoMethodError` while trying to perform the [following code](https://github.com/oauth-xx/oauth2/blob/c3c49a05ad17bb89aa43df35ec66e30f523667ae/lib/oauth2/client.rb#L225): `response.parsed.merge(access_token_opts)`, because the `response.parsed` is nil.

pboling:
Great analysis.  Do you think you could work up a failing spec in a PR?

joaolrpaulo:
Already did, #460

joaolrpaulo:
Already closed for sometime

PR #460, opened 2019-07-09
Fix client `fetch_token` throwing `NoMethodError` when raise_errors is set to false

joaolrpaulo:
The problem occurs when the identifier service is unreachable and we are passing the :raise_errors option set to false.

The `fetch_token` method throws `NoMethodError` while trying to perform the following code:

`response.parsed.merge(access_token_opts)`

This happens because the `response.parsed` is `nil`.

### DIFF
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -149,7 +149,10 @@ module OAuth2
       if options[:raise_errors] && !response_contains_token
         error = Error.new(response)
         raise(error)
+      elsif !response_contains_token
+        return nil
       end
+
       build_access_token(response, access_token_opts, access_token_class)
     end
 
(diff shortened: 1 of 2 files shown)


joaolrpaulo:
Related to opened issue: #459

joaolrpaulo:
Hello @pboling, I was really short on time to contribute to this PR. But here you go.

My suggestion to this is to return nil when there is now way we can build the access token, which is the saying the same that `!response_contains_token`.

Cheers mate!

pboling:
Thanks @joaolrpaulo !!

joaolrpaulo:
Hello @pboling, is there a possibility to add this to a new patch release in order for we to use it ?

pboling:
@joaolrpaulo it would have to be backported to the [1-4-stable](https://github.com/oauth-xx/oauth2/tree/1-4-stable) branch, as a second PR.

joaolrpaulo:
I will do it in the next few days :pray:

Issue #386, opened 2018-06-03
Path Stripped From Site


twodee:
Hi, I've got a similar issue to #245, but the fix there doesn't apply. I'm trying to use oauth2 with WordPress, and I'm getting broken redirects. The issue is that oauth2 considers only the root of the site when generating URLs, stripping out all path information. This leads to incompatibilities with tools like WordPress that allow for more flexible notions of sites.

On the server side, the blog I am trying to add OAuth2 authentication to resides at (fictional) URL https://somedomain.org/blog. I am using WP OAuth Server, a WordPress plugin available from https://wp-oauth.com. It considers the site to be https://somedomain.org/blog, because that's the URL that WordPress gives it.

On the client side, I configure the authorization in the following manner:

````
oauthor = OAuth2::Client.new(CLIENT_ID, CLIENT_SECRET, {
  site: 'https://somedomain.org',
  authorize_url: '/blog/oauth/authorize',
  token_url: '/blog/oauth/token',
  redirect_uri: REDIRECT_URI
})
````

When I visit the URL generated by `oauthor.auth_code.authorize_url`, it redirects to the WordPress login form, but it passes `https://somedomain.org/blog/blog/oauth/authorize` as the URL to redirect to on successful authorization. The doubled up "blog" causes a 404.

It would seem to me that there should be some way for oauth2 to work with a site that doesn't reside at the root. That way it can work with servers like WordPress that operate in a more relative way. This problem would be fixed if I could use this configuration:

````
oauthor = OAuth2::Client.new(CLIENT_ID, CLIENT_SECRET, {
  site: 'https://somedomain.org/blog',
  authorize_url: '/oauth/authorize',
  token_url: '/oauth/token',
  redirect_uri: REDIRECT_URI
})
````

Any help is appreciated.

cyruscavalcante:
+1

elliotcm:
@twodee I've looked into this for you (see https://github.com/oauth-xx/oauth2/pull/461) and it looks like you can get the behaviour you want by dropping the leading `/` on your `authorize_url` and `token_url` options.

Issue #245, opened 2016-03-22
Requests are made against site root, not site


jensljungblad:
If I set the site root to: `https://www.example.com/foobar`

Requests are posted to `https://www.example.com/oauth/token` and not `https://www.example.com/foobar/oauth/token` which I would expect. How come?

My problem is that I'm trying to authenticate against a system which uses the following URL:s for staging and production respectively:

```
https://test.example.com/login
https://login.example.com
```

jasper-lyons:
#241 seems to cover this?

Try `authorize_url: '/foobar'` and see if you get any success.

xtagon:
Closing due to inactivity - if the `authorize_url` param didn't solve this we can re-open.

pboling:
Fixed by #461

PR #461, opened 2019-07-15
Add spec for sites with path components

elliotcm:
This was raised in issues 245 and 386, the latter specificially regarding wordpress blogs which often define a "site" as including a path component.

This behaviour is already supported by Faraday (the underlying HTTP client) so long as you specify relative paths when constructing the new URLs rather than absolute ones, e.g. "oauth/authorize" will work but "oauth/token" will not.

It might be desirable to handle this automatically for users by stripping leading `/` characters from the authorize and token URL options (which should really be called `*_path`?), but could also be handled through documentation.

This commit adds a spec demonstrating the use of relative paths as per issue 386.

https://github.com/oauth-xx/oauth2/issues/386
https://github.com/oauth-xx/oauth2/issues/245

### DIFF
--- a/spec/oauth2/client_spec.rb
+++ b/spec/oauth2/client_spec.rb
@@ -4,7 +4,7 @@ require 'nkf'
 
 RSpec.describe OAuth2::Client do
   subject do
-    described_class.new('abc', 'def', :site => 'https://api.example.com') do |builder|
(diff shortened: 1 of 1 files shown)


elliotcm:
@pboling Hi Peter.  I've had a poke through your 2.0.0 milestone issues and the one mentioned here seems to be the only one not already handled in some way.

pboling:
@elliotcm Thank you `x` 💯

PR #464, opened 2019-07-20
Allow use faraday 1.0

Fudoshiki:


etiennebarrie:
Would it be possible to release a 1.x version (i.e. `v1.4.2`) with this included? I don't think we need the rest of the changes from 2.0 to be able to loosen that constraint.

pboling:
Just need a PR to https://github.com/oauth-xx/oauth2/tree/1-4-stable, that's why we made the branch!  My fear is that it would be horribly broken... but let's find out.

etiennebarrie:
I just saw v1.4.2, that's awesome. Thank you very much!

pboling:
It was broken, just in the specs, but I fixed it all up!

CC @Fudoshiki

Issue #193, opened 2014-09-12
Client secret found in GET URI MUST not exist


maletor:
“The parameters can only be transmitted in the request-body and MUST NOT be included in the request URI.”

From the OAuth spec, we should not allow client_secret to ever appear in the URL, yet there is [a spec for it](https://github.com/intridea/oauth2/blob/master/spec/oauth2/strategy/auth_code_spec.rb#L12).

pboling:
Fix will be release in v2.0.0

PR #472, opened 2019-09-20
Add checks to enforce no client_secret in authorize_url query params

dfockler:
Fix for #193 

Raise error if the `implicit`, and `auth_code` grant types are passed the `client_secret` as a parameter for the `authorize_url`.


pboling:
Merging this in spite of some of the build envs being in a broken state.  Most of the builds are passing, and this change wouldn't result in any of the failures seen.  We need to restrict the gem versions (particularly of bundler) per each version of Ruby.

PR #485, opened 2019-12-18
Add the ability to use TLS client auth when requesting access token

thomcorley:
When making the request to retrieve an access token, if the API requires
you to use TLS client auth as the authentication mechanism then it will
fail if the client secret is sent. For this auth type, only the client
ID should be sent.

We do not currently allow the developer to customise this; we set the
client ID and secret by default for all access token requests.

This PR adds an extra section to the case statement in
Authenticator#apply to add the desired functionality; when passing in
`tls_client_auth` as the auth mode, the client secret will be omitted
from the request.

**Update:**
There were some failing specs due to the logging in the new version of Faraday (1v1.0.0) being in a different format to that which the test suite expected. This PR also addresses this. The fix was to remove the log level markers "INFO/DEBUG".

### DIFF
--- a/lib/oauth2/authenticator.rb
+++ b/lib/oauth2/authenticator.rb
@@ -25,6 +25,8 @@ module OAuth2
         apply_basic_auth(params)
       when :request_body
         apply_params_auth(params)
+      when :tls_client_auth
+        apply_tls_params_auth(params)
       else
         raise NotImplementedError
       end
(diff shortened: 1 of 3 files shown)


thomcorley:
@pboling are you a maintainer of this gem? I wonder if you might be able to find some time to review this small PR? I work in the banking team at FreeAgent and we want to be able to use `tls_client_auth` as an auth scheme for making Open Banking API connections. It would be great if this change was pushed to master!

By the way - I encountered an issue with the logging from Faraday that was causing the test suite to fail, and this PR fixes that also. My colleague @dgholz also put up a PR to fix the specs:
https://github.com/oauth-xx/oauth2/pull/493

pboling:
@thomcorley Yes I am the primary maintainer!  Sorry I missed this PR compeltely over the holidays.  I'll get it into master (tracking the future 2.0 release), and try to backport to 1.4.x

thomcorley:
Thank you for the prompt review @pboling! This is going to be a great help for us.

pboling:
This was released in 1.4.3+, and will also be in 2.x series.

Issue #486, opened 2019-12-19
301 Redirect should handle relative URL in response_headers['location']


cowlibob:
The returned location is used for the next request directly, but in some cases, this can be a relative URL:

> An obsolete version of the HTTP 1.1 specifications (IETF RFC 2616) required a complete absolute URI for redirection.[2] The IETF HTTP working group found that the most popular web browsers tolerate the passing of a relative URL[3] and, consequently, the updated HTTP 1.1 specifications (IETF RFC 7231) relaxed the original constraint, allowing the use of relative URLs in Location headers.[4]
 (https://en.wikipedia.org/wiki/HTTP_location)

https://github.com/oauth-xx/oauth2/blob/3bef9feff5de9d46b35c7423061508e5092b1eac/lib/oauth2/client.rb#L113

cowlibob:
I think the best solution is to merge the `response.headers['location']` value into the original URL `response.response.url`.

`response.response.env.url.merge (response.headers['location'])`

pboling:
Ref #487

PR #487, opened 2019-12-19
Handle relative location in a 301 redirect.

cowlibob:
This is a simple change to merge a relative reference (1) or absolute URI (2) from the location header into the original request URI.

If there are any barriers to this being merged, please let me know.

1. https://tools.ietf.org/html/rfc3986#section-4.2
2. https://tools.ietf.org/html/rfc3986#section-4.3

### DIFF
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -110,7 +110,8 @@ module OAuth2
           verb = :get
           opts.delete(:body)
         end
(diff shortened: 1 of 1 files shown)


pboling:
@cowlibob Please rebase on top of latest master!

pboling:
There should be a spec added to test this as well.

pboling:
Fixed by #486 and #575

PR #494, opened 2020-01-27
Adds support for private_key_jwt

SteveyblamFreeagent:
A recent merge added support for tls_client_auth scheme by providing an auth scheme value that would only add the client_id to the request body. The Private Key JWT scheme includes the client_id in the client_assertion and so is not needed in the main request body. 

This PR simply returns unaltered params when private_key_jwt is specified as the auth_scheme

### DIFF
--- a/lib/oauth2/authenticator.rb
+++ b/lib/oauth2/authenticator.rb
@@ -25,8 +25,8 @@ module OAuth2
         apply_basic_auth(params)
       when :request_body
         apply_params_auth(params)
-      when :tls_client_auth
-        apply_tls_params_auth(params)
+      when :tls_client_auth, :private_key_jwt
+        apply_client_id(params)
       else
         raise NotImplementedError
       end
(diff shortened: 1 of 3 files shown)


lib/oauth2/authenticator.rb
```diff
@@ -44,8 +46,9 @@ def apply_params_auth(params)
       {'client_id' => id, 'client_secret' => secret}.merge(params)
     end
 
-    # When using the TLS client auth scheme, we don't want to send the secret
-    def apply_tls_params_auth(params)
+    # When using schemes that don't require the client_secret to be passed ( TLS Client auth, Private Key JWT etc),
```
pboling: Same problem with the comment as in the other PR.

pboling: I'll merge this and then make the change.

pboling:
Based on OAuth JWT assertion specification [(RFC 7523)](https://tools.ietf.org/html/rfc7523)

PR #495, opened 2020-01-27
add support for private_key_jwt and tls_client_auth

SteveyblamFreeagent:
A recent merge to master added support for tls_client_auth scheme by providing an auth scheme value that would only add the client_id to the request body. The Private Key JWT scheme includes the client_id in the client_assertion and so is not needed in the main request body.

This PR simply returns unaltered params when private_key_jwt is specified as the auth_scheme and adds tls support for v1.4

I have also dropped the INFO/DEBUG from the client spec as it appears to switch between the two depending on the ruby version. I thought this was neater and simpler than writing a complex set of contexts for different versions. Especially considering that the important aspect is the output information rather than the label.

### DIFF
--- a/lib/oauth2/authenticator.rb
+++ b/lib/oauth2/authenticator.rb
@@ -25,6 +25,10 @@ module OAuth2
         apply_basic_auth(params)
       when :request_body
         apply_params_auth(params)
+      when :tls_client_auth
+        apply_client_id(params)
+      when :private_key_jwt
+        params
       else
         raise NotImplementedError
       end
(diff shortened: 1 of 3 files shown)


lib/oauth2/authenticator.rb
```diff
@@ -25,6 +25,10 @@ def apply(params)
         apply_basic_auth(params)
       when :request_body
         apply_params_auth(params)
+      when :tls_client_auth
+        apply_client_id(params)
+      when :private_key_jwt
+        params
```
pboling: @SteveyblamFreeagent just straight `params`?  The comment on line 49 appears to imply that `apply_client_id` would be used for `private_key_jwt` also.

SteveyblamFreeagent: Thanks for looking at this Peter. Straight params is expected behaviour as private_key_jwt wraps up the client_id in the assertion so it's not passed in the request body. But you're absolutely right the comment was from a previous commit where I was applying the client_id. I'll change the comment now.

Many thanks

PR #507, opened 2020-06-05
Fix camel case content type, response keys

anvox:
Changes:

- Downcase content type before use. It'll less surprise than just downcase when used.
- Snake case body keys
- Remove deprecate warn on test

Replace https://github.com/oauth-xx/oauth2/pull/202 as I see it isn't updated for a long time.

### DIFF
--- a/lib/oauth2/response.rb
+++ b/lib/oauth2/response.rb
@@ -79,7 +79,7 @@ module OAuth2
           end
         end
 
-      @parsed = OAuth2::SnakeHash.build(@parsed) if @parsed.is_a?(Hash)
+      @parsed = OAuth2::SnakyHash.build(@parsed) if @parsed.is_a?(Hash)
 
       @parsed
     end
--- a/lib/oauth2/response.rb
+++ b/lib/oauth2/response.rb
@@ -92,7 +92,7 @@ module OAuth2
     # Attempts to determine the content type of the response.
     def content_type
       return nil unless response.headers
-      ((response.headers.values_at('content-type', 'Content-Type').compact.first || '').split(';').first || '').strip
+      ((response.headers.values_at('content-type', 'Content-Type').compact.first || '').split(';').first || '').strip.downcase
     end
 
     # Determines the parser (a Proc or other Object which responds to #call)
@@ -121,7 +121,7 @@ module OAuth2
           @@parsers[options[:parse].to_sym]
         end
 
-      @parser ||= @@parsers[@@content_types[content_type.downcase]]
+      @parser ||= @@parsers[@@content_types[content_type]]
     end
 
     private
--- a/lib/oauth2/response.rb
+++ b/lib/oauth2/response.rb
@@ -79,12 +79,7 @@ module OAuth2
           end
         end
 
-      if @parsed.is_a?(Hash)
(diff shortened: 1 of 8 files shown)


anvox:
@pboling Could you help to review? Thanks.

lib/oauth2/response.rb
```diff
@@ -78,12 +78,21 @@ def parsed
             parser.call(body, response)
           end
         end
+
+      if @parsed.is_a?(Hash)
```
pboling, line 82: Would like to remove the internal handling, and use a tool that gives us more coverage, such as the ability to snake case the hashes inside JSON **arrays** as well.

[plissken](https://github.com/futurechimp/plissken) looks like a good candidate.

lib/oauth2/response.rb
```diff
@@ -114,6 +123,15 @@ def parser
...
       @parser ||= @@parsers[@@content_types[content_type]]
     end
+
+    private
+
+    def to_snake_case(string)
+      string.gsub(/([A-Z]+)([A-Z][a-z])/, '\1_\2')
+            .gsub(/([a-z\d])([A-Z])/, '\1_\2')
+            .tr("-", "_")
+            .downcase
```
pboling, line 133: The snake casing in [`plissken `](https://github.com/futurechimp/plissken/blob/master/lib/plissken/methods.rb) is a bit more robust.
```
        string.tr("::", "/")
              .gsub(/([A-Z]+)([A-Z][a-z])/, '\1_\2')
              .gsub(/([a-z\d])([A-Z])/, '\1_\2')
              .tr("-", "_")
              .downcase
```
Let's switch to **plissken** and remove this internal snake case logic.  We'll get much more robust testing as a result.

pboling:
@anvox I like this idea!

Would need to be a part of a major version bump.  Also would be nice to somehow keep a reference to the original keys, because there may be some tools that for whatever reason, need to access the keys as-is, rather than snake cased.

I have an idea for how that could work...  Need some time to sort it out.

lib/oauth2/response.rb
anvox: Thanks. Agree that `plissken` changes are to widely. 
I think this code, which from #202, is original from `plissken`. I remove `tr("::", "/")` because I think it's mostly use to transform full class name to snake case. In our case, I believe we don't have `::` in header name.

anvox:
I'm thinking about a custom type similar to indifferent hash, but prefer to hear you first. 
Meanwhile, I'm looking into [2.0.0 release milestone](https://github.com/oauth-xx/oauth2/milestone/1)

pboling:
Yes, exactly.  Like an indifferent hash, but actually just an alternate key map.

lib/oauth2/snake_hash.rb
```diff
@@ -0,0 +1,35 @@
+module OAuth2
+  class SnakeHash < ::Hash
```
anvox, line 2: @pboling I tried this approach https://gist.github.com/anvox/c4e2ca33c6fa59d786cf97960157ae9b but it's overkill. Need to rewrite most methods like indifferent hash. I think just a hash which ref snake case key to camel case key is good enough.

PR #196, opened 2014-09-16
Update MAC to draft 05

singpolyma:
Closes #194

Issue #194, opened 2014-09-15
MAC timestamp should be in milliseconds now


singpolyma:
So says http://tools.ietf.org/html/draft-ietf-oauth-v2-http-mac-05

pboling:
This was fixed by #509

pboling:
MAC Token support is being dropped from the OAuth2 gem.

See #589

PR #509, opened 2020-06-11
Update Mac to draft 05

anvox:
Remake https://github.com/oauth-xx/oauth2/pull/196 with a little change in detail. 
See comments. 

Implement https://github.com/oauth-xx/oauth2/issues/194

### DIFF
--- a/lib/oauth2/mac_token.rb
+++ b/lib/oauth2/mac_token.rb
@@ -29,6 +29,9 @@ module OAuth2
     # @option opts [FixNum, String] :algorithm (hmac-sha-256) the algorithm to use for the HMAC digest (one of 'hmac-sha-256', 'hmac-sha-1')
     def initialize(client, token, secret, opts = {})
       @secret = secret
+      @seq_nr = SecureRandom.random_number(2 ** 64)
+      @kid = opts.delete(:kid) || Base64.strict_encode64(Digest::SHA1.digest(token))
+
       self.algorithm = opts.delete(:algorithm) || 'hmac-sha-256'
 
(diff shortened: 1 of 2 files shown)


anvox:
I see https://github.com/oauth-xx/oauth2/pull/196 inactive for a long time, so I remake.

lib/oauth2/mac_token.rb
```diff
@@ -29,6 +29,9 @@ def self.from_access_token(token, secret, options = {})
     # @option opts [FixNum, String] :algorithm (hmac-sha-256) the algorithm to use for the HMAC digest (one of 'hmac-sha-256', 'hmac-sha-1')
     def initialize(client, token, secret, opts = {})
       @secret = secret
+      @seq_nr = SecureRandom.random_number(2 ** 64)
```
anvox, line 32: > Sequence numbers fall in the range 0 through 2^64 - 1 and wrap to zero following the value 2^64 - 1.

As [SecureRandom.random_number(n)](https://apidock.com/ruby/SecureRandom/random_number/class) generate number => 0 and < n, we needn'nt minus 1

pboling:
@anvox This looks great!  This appears to be a breaking change?  Would you suggest a minor release, or a major release?  How "breaking" is this change?

anvox:
@pboling I think it should be a major change. By this change, we no longer support MAC token draft < 3. Clients must change from using nonce to seq-nr.

pboling:
@anvox ok, this will be a new major release of 2.0.0, and the old major milestones will all get bumped up by +1.0.0

Issue #123, opened 2012-12-15
AccessToken from_hash method makes access token location assumption


brittballard:
The AccessToken from_hash method assumes the access_token is at the root level of the json response. It also assumes the key is "access_token". This isn't always the case.

pboling:
@gpinkham , yes it is super easy to do a bad implementation of OAuth2. 👎 

Closing.

PR #511, opened 2020-06-25
Ability to retrieve access token string from within nested data structure

Capncavedan:
Updated to work with rubies prior to 2.3 (which lack `Hash#dig`) 

I have a need to work with an Oauth provider that does not conform to the Oauth standard for how the access token is returned.

Pinterest Oauth for example at this time returns e.g.
```json
{ "data": { "access_token": "something" } }
```

when the Oauth spec wants `access_token` to be at the top level.

This PR allows configuring the client with:
  `dig_nested_keys_for_access_token: "data,access_token"`

which will allow "something" to be fetched from the above example.

I have researched other solutions and ran across [this suggestion](https://github.com/oauth-xx/oauth2/issues/123#issuecomment-11398046) from several years ago, but given how the code is structured today, subclassing AccessToken and modifying its behavior would not work - as the exception is raised [here](https://github.com/oauth-xx/oauth2/blob/e86358e6043b6c1f5f65583a4afd913123828341/lib/oauth2/client.rb#L150) before the access token object is built [here](https://github.com/oauth-xx/oauth2/blob/e86358e6043b6c1f5f65583a4afd913123828341/lib/oauth2/client.rb#L157).

### DIFF
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -150,11 +150,8 @@ module OAuth2
       response_contains_token = response.parsed.is_a?(Hash) &&
                                 (response.parsed['access_token'] || response.parsed['id_token'])
 
-      if response.parsed.is_a?(Hash) && options[:dig_nested_keys_for_access_token]
-        if nested_access_token = response.parsed.dig(*options[:dig_nested_keys_for_access_token].to_s.split(","))
-          nested_access_token_hash = { "access_token" => nested_access_token }
-          response_contains_token = true
-        end
+      if nested_access_token_hash = dig_response(response, options[:dig_nested_keys_for_access_token])
+        response_contains_token = true
       end
 
       if options[:raise_errors] && !response_contains_token
@@ -244,5 +241,22 @@ module OAuth2
     def oauth_debug_logging(builder)
       builder.response :logger, options[:logger], :bodies => true if ENV['OAUTH_DEBUG'] == 'true'
     end
+
+    def dig_response(response, dig_string)
+      return nil unless response.parsed.is_a?(Hash)
+      return nil unless dig_string.is_a?(String)
+
+      sub_element = response.parsed.dup
+      access_token_string = nil
+      dig_string.split(",").each do |key|
+        sub_element = sub_element[key]
+        access_token_string = sub_element
+        break if sub_element.nil?
+      end
+
+      return nil if access_token_string.nil?
+      { "access_token" => access_token_string }
+    end
+
   end
 end
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -26,6 +26,8 @@ module OAuth2
     # @option opts [Boolean] :raise_errors (true) whether or not to raise an OAuth2::Error
     # @options opts [Logger] :logger (::Logger.new($stdout)) which logger to use when OAUTH_DEBUG is enabled
     #  on responses with 400+ status codes
+    # @options opts [String] :dig_nested_keys_for_access_token comma-separated list of attributes to dig within the response hash to
+    #  retrieve the Oauth access token. e.g. for Pinterest API v3: "data,access_token"
     # @yield [builder] The Faraday connection builder
     def initialize(client_id, client_secret, options = {}, &block)
       opts = options.dup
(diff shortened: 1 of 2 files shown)


pboling:
@Capncavedan I just wonder if we expand this slightly, me might be able to repeat/remap any value that providers may have placed in odd spots.

Capncavedan:
@pboling that seems possible .... perhaps a hash with the attribute to find as the key and the path to it as the key?  A (made up) example to fetch access token from `data ==> access_token` and uid from `data ==> user ==> id`

```ruby
{
  "access_token" => "data,access_token",
  "uid" => "data,user,id"
}
```

is that along the lines of what you were thinking, or did I misunderstand?

pboling:
@Capncavedan Yes, that is it exactly.  There are many ouath providers that are non-standard in different ways.

Capncavedan:
@pboling got it, and agreed - many non-standard providers out there.  Do you think proceed with this PR and pursue that separately, or would rather only do one that's a more generic solution?

anvox:
@Capncavedan I think we could make it in this PR to make both access token and uid use the same dig mechanism. 
How about key `access_token_path`? I just think about it because it's shorter :D

pboling:
@Capncavedan I'd like to go with a more generic solution, as you suggested here:
```
{
  "access_token" => "data,access_token",
  "uid" => "data,user,id"
}
```
@anvox I think you missed some of the comments. We are working toward a more generic solution, nothing with `*_path`.

Additionally, the `master` branch is targeting a future major release, probably 3.0.0 at this point. This is a non-breaking change and could certainly fit in a patch release on the 1.4.x line.  You'll need to use `1-4-stable` branch for that.  Of course we *also* need the work to be done in master.

anvox:
Ah, see it :) Thanks.

lib/oauth2/client.rb
```diff
@@ -26,6 +26,8 @@ class Client # rubocop:disable Metrics/ClassLength
     # @option opts [Boolean] :raise_errors (true) whether or not to raise an OAuth2::Error
     # @options opts [Logger] :logger (::Logger.new($stdout)) which logger to use when OAUTH_DEBUG is enabled
     #  on responses with 400+ status codes
+    # @options opts [String] :dig_nested_keys_for_access_token comma-separated list of attributes to dig within the response hash to
+    #  retrieve the Oauth access token. e.g. for Pinterest API v3: "data,access_token"
```
jonspalmer, line 30: Is there a reasons to make this a comma delimited string vs just an array of strings?

lib/oauth2/client.rb
```diff
@@ -143,18 +145,23 @@ def get_token(params, access_token_opts = {}, access_token_class = AccessToken)
         opts[:headers] = {}
       end
       opts[:headers].merge!(headers)
+
       response = request(options[:token_method], token_url, opts)
       response_contains_token = response.parsed.is_a?(Hash) &&
                                 (response.parsed['access_token'] || response.parsed['id_token'])
 
+      if nested_access_token_hash = dig_response(response, options[:dig_nested_keys_for_access_token])
+        response_contains_token = true
```
jonspalmer, line 154: Shouldn't this only set `response_contains_token` to be true if `nested_access_token_hash` actually exists?

jonspalmer:
I came across this issue researching a related issue with Azure B2C. There the issue isn't nesting its that sometimes only id_token is present and not access_token. My sense is it would be more powerful to solve these problems by making the subclassing AccessToken process work (which it doesn't really today). Thoughts?

lib/oauth2/client.rb
jonspalmer: Can we use this solution to solve the case above which exists to solve the problem that Microsoft implementations only supply `id_token` not `access_token` when only then openid scope is requested. 

That case is a bit more complicated in that it wants `access_token` if it exists otherwise `id_token`. It also exposes the problem that you identity here with is that pulling out the token is smeared across a few classes. Presumably you're doing the merge later so you don't have to subclass AccessToken, reimplementing `from_hash` to get the nested access token.

How about we just generalize the concept with a client option proc called `extract_access_token` by default it would be `Proc.new { |response|  response.parsed['access_token'] }`. Microsoft clients would be `Proc.new { |response|  response.parsed['access_token'] || response.parsed['id_token'] }` and your use case  `Proc.new { |response| response.parsed.dig(:data,:access_token) }`

jonspalmer:
@anvox I have some ideas for a different shape. The problem as I see it right now is that `OAuth2::Client` is trying determine valid responses but is still deferring the actual extraction of the token the the AccessToken class. My idea would be to do all of that in `from_hash`. We always call `build_access_token`. Iff the access token class doesn't find a token of its liking then `from_hash` should return nil and we could raise errors based on that. I'll throw together a PR.

I don't have docs of the Microsoft issue. However, its provable calling their APIs 😄 v1 returns both access_token and id_token v2 only id_token. It's the source of https://github.com/Meeco/omniauth-azure_active_directory_b2c/issues/2 so I'm not the only one seeing it.

lib/oauth2/client.rb
pboling: That is an excellent idea @jonspalmer

pboling:
@jonspalmer @anvox After merging #518 we no longer need this open, right?

anvox:
Yeah, #518 covered this function.

Issue #498, opened 2020-02-28
Cannot provide custom headers to get_token request 


antonlitvinenko:
Hello!

I think this change https://github.com/oauth-xx/oauth2/commit/48cd561bb3a3c8d10d2ad5d07b5fc765cb32f216 has introduced a bug which prevents from setting custom headers on get_token request.

1. Consider code `client.auth_code.get_token(verifier,{headers: ['Header 1' => 'A']}, ...)`
2. When it reaches the `AuthCode.get_token` then every key in `params` is rewritten be a string. So, on when calling `@client.get_token(params, opts)` the params are changed to `{'headers' => ['Header 1' => 'A']}`
3. `OAuth2::Client.get_token` there is a line which seems to take custom headers and apply them to the request, except it accesses headers using symbol: `headers = params.delete(:headers) || {}` , but AuthCode has rewritten all the keys to be strings. So, `params.delete(:headers)` returns `nil`

This seems like a big issue and someone should have already noticed it since the commit. But I couldn't find anybody mentioning it, so I have a feeling that I am missing something and this code, in fact, is working as expected.

pboling:
We need to write specs around this scenario if there are none.  Thanks for discovering it.  Need to think about how to handle string/symbol more carefully, and consistently across the codebase.

pboling:
@anvox Did you check for symbol key access in the code when you did the SnakeyHash work?  I think we need to double check for symbol access and figure out how to deal with it.

anvox:
@pboling  Currently, SnakyHash only checks string key. 
Let me write a test and test against `HashWithIndifferentAccess`, I think it fits this case.

pboling:
@anvox FYI, I had to switch to a more robust implementation of snaky hash - https://github.com/shishi/rash_alt

@antonlitvinenko This issue should be resolved in `master` branch once #575 is merged.

PR #512, opened 2020-06-28
Fix missing custom headers in get_token

anvox:
I propose this way to quickly get over #498 
If we use `HashWithIndifferentAccess` or anything similar, we must revise and change whole system inputs, outputs. Currently, system returns hashes with mixed keys in string and symbol. 
This change looks like a patch but it has 2 pros besides fixing issue. 

- Simple implement, we needn't add complex things like HashWithIndifferentAccess
- Keep others interfaces same. 

After, we could have a bigger implement to standardize inputs, outputs.


anvox:
@pboling  @antonlitvinenko  what do you think?

pboling:
This looks good.  This bug is in the current release also, so we'll need another PR to get this same code into `1-4-stable`.  Can you do that as well @anvox ?  I think the simplest may be to squash merge this PR, and then take that squash merged commit SHA and cherry pick it into a branch off of `1-4-stable` for a new PR.

Going to squash merge this into master now.

antonlitvinenko:
@anvox awesome! Thank you very much!

anvox:
Thanks. PR to 1-4-stable https://github.com/oauth-xx/oauth2/pull/513 
I'll merge after all tests green.

PR #514, opened 2020-07-06
add token in the get_token method

Alice-Qiu:


anvox:
Thanks @Alice-Qiu, we're going to make it configurable in https://github.com/oauth-xx/oauth2/pull/511

PR #516, opened 2020-07-13
Add Client option for access_token_class

jonspalmer:
## The Problem

There is no way to use `OAuth2::Client` strategies such as `auth_code` AND supply a custom access_token_class.

`OAuth2::Client.get_token` has an optional argument for the access_token_class to instantiate. 
However, callers of it have to know to use it and none of the strategies do so. This means the anyone needing a custom access_token_class, #243 or #511, need to jump through hoops to get the behavior they want. For example subclassing `OAuth2::Client` just to change the default class

## What I did

Added a new option, `access_token_class`, to `OAuth2::Client` which is used as the default `access_token_class` for all calls to `get_token`.

This is a safe, non-breaking change.

### DIFF
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -28,6 +28,7 @@ module OAuth2
     # @option options [Boolean] :raise_errors (true) whether or not to raise an OAuth2::Error
     # @option options [Logger] :logger (::Logger.new($stdout)) which logger to use when OAUTH_DEBUG is enabled
     #  on responses with 400+ status codes
+    # @option options [Class] :access_token_class (AccessToken) class used to create access tokens
     # @yield [builder] The Faraday connection builder
     def initialize(client_id, client_secret, options = {}, &block)
       opts = options.dup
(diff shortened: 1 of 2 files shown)


pboling:
Would it be too complex to bother trying this on the `1-4-stable` branch?

jonspalmer:
> @jonspalmer Would it be too complex to bother trying this on the `1-4-stable` branch?

@pboling #518 provides a better solution to this problem and is already in `1-4-stable`. That change is going to merge conflict with this change in master and IMO should replace this change entirely. The point being that the proc approach in #518  allows you to return any flavor of AccessToken you desire.

pboling:
@jonspalmer Thanks for the heads up, again, and sorry for the long delay in addressing it.  v2 is moving ahead at a very slow pace because my personal use case for this gem is vanishingly thin, and I would need far more community support to move it forward more quickly.

I'm going to work on parity between 1-4-stable and v2 (master, soon to be main) now.

PR #517, opened 2020-07-13
Defer parsing and validation of access tokens to the AccessToken class

jonspalmer:
This is an alternative take on a solution for #511

Needs #516 to be able to fix the failing specs and add more spec coverage.

With this fix in place the solution to #511 would be to subclass `AccessToken`. Something like this:

```ruby
class PinterestAccessToken < OAuth2::AccessToken
  class << self
    def from_hash(client, hash)
      hash = hash.dup
      token =  hash.dig(:data, :access_token)
      return unless token
      new(client, token, hash)
    end
  end

  #... any class level changes
end
```

Similarly the Microsoft id_token only problem could be solved with

```ruby
class MicrosoftAccessToken < OAuth2::AccessToken
  class << self
    def from_hash(client, hash)
      hash = hash.dup
      token =  hash.delete("access_token") || hash.delete("id_token")
      return unless token
      new(client, token, hash)
    end
  end
end
```

This is likely a breaking change as it requires a new API for anyone that has subclassed `AccessToken`.

### DIFF
--- a/lib/oauth2/access_token.rb
+++ b/lib/oauth2/access_token.rb
@@ -11,7 +11,9 @@ module OAuth2
       # @return [AccessToken] the initalized AccessToken
       def from_hash(client, hash)
         hash = hash.dup
-        new(client, hash.delete('access_token') || hash.delete(:access_token), hash)
+        token =  hash.delete('access_token') || hash.delete(:access_token)
+        return unless token
+        new(client, token, hash)
       end
 
       # Initializes an AccessToken from a key/value application/x-www-form-urlencoded string
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -158,17 +158,15 @@ module OAuth2
       end
       opts[:headers].merge!(headers)
       response = request(options[:token_method], token_url, opts)
-      response_contains_token = response.parsed.is_a?(Hash) &&
-                                (response.parsed['access_token'] || response.parsed['id_token'])
 
-      if options[:raise_errors] && !response_contains_token
+      access_token = build_access_token(response, access_token_opts, access_token_class)
+
+      if options[:raise_errors] && !access_token
         error = Error.new(response)
         raise(error)
-      elsif !response_contains_token
-        return nil
       end
 
-      build_access_token(response, access_token_opts, access_token_class)
+      access_token
     end
 
     # The Authorization Code strategy
@@ -240,6 +238,7 @@ module OAuth2
     #
     # @return [AccessToken] the initialized AccessToken
     def build_access_token(response, access_token_opts, access_token_class)
+      return unless response.parsed.is_a?(Hash)
       access_token_class.from_hash(self, response.parsed.merge(access_token_opts)).tap do |access_token|
         access_token.response = response if access_token.respond_to?(:response=)
       end
--- a/spec/oauth2/client_spec.rb
+++ b/spec/oauth2/client_spec.rb
@@ -463,29 +463,51 @@ RSpec.describe OAuth2::Client do
     end
 
     context 'when the :raise_errors flag is set to false' do
-      context 'when the request body is nil' do
-        it 'returns a nil :access_token' do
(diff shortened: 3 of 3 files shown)


anvox:
@jonspalmer  Could you help to merge https://github.com/oauth-xx/oauth2/pull/516 to this PR? I think they bring better value if went together. This improvement is the best use of access token class option per client. 

I don't have any problem with this approach, just some concerns which makes me prefer passing mapping options to guide client extracts access token than. 

- This is a big change, which may hard to merge back to 1-4-stable
- Must extend `AccessToken` to use with vary providers. To extend, users must understand `AccessToken`, which I think take more time than using an option. 
- Should we provide common access token like `PinterestAccessToken`, `MicrosoftAccessToken` or just documentation and leave use use it?

lib/oauth2/client.rb
```diff
@@ -158,17 +158,15 @@ def get_token(params, access_token_opts = {}, access_token_class = AccessToken)
       end
       opts[:headers].merge!(headers)
       response = request(options[:token_method], token_url, opts)
-      response_contains_token = response.parsed.is_a?(Hash) &&
-                                (response.parsed['access_token'] || response.parsed['id_token'])
 
-      if options[:raise_errors] && !response_contains_token
+      access_token = build_access_token(response, access_token_opts, access_token_class)
```
anvox, line 162: 👍 Determine response_contains_token or token valid should be token's responsible.

jonspalmer:
@anvox I agree they work well together but I'm not convinced we should merge things. #516 is a completely safe, non-breaking change that we can merge into 1-4-stable with no major issues. As you point out this is a bigger change and might not be appropriate for 1-4-stable. #516 would at least make life a bit easier for consumers trying to solve problems that this PR is aiming at.

To address your questions:
 * I'm not %100 sure on the status of stable branches vs 2.0 etc. what the timelines and approaches are
 * We really don't need people to know everything about AccessToken if mostly just `from_hash` that matters. Frankly that pattern itself is super weird. It would be better if everything was is `OAuth2::Client`
 * Doc - No I don't think we should add all the example implementations. Document it well and let others implement them. The  popular combinations can become their own gems.

lib/oauth2/access_token.rb
```diff
@@ -11,7 +11,9 @@ class << self
       # @return [AccessToken] the initalized AccessToken
       def from_hash(client, hash)
         hash = hash.dup
-        new(client, hash.delete('access_token') || hash.delete(:access_token), hash)
+        token =  hash.delete('access_token') || hash.delete(:access_token)
```
anvox, line 14: I'm thinking at this point, maybe we use an option from client like https://github.com/oauth-xx/oauth2/pull/511/files#diff-bc9331245a41740ee0145e2d5640ed07R29 but not limit to 1 but multiple mapping. i.e. 

```ruby
{
  access_token: ['access_token', 'id_token']
}
```

Then try each mapping. 
For Pinterest
```ruby
{
  access_token: 'data,access_token' # OR access_token: ['data,access_token']
}
```

Then 
```ruby
token = nil
client.response_mapping[:access_token].detect do |mapping|
  token = hash.dig(mapping.split(','))
end
```

What do you think? @jonspalmer

jonspalmer: I don't think a single option for the "path" to dig is enough. Not only do you want to get the access token but you also need to remove it from the hash before actually constructing the `AccessToken` etc. Various customized clients might have reason to otherwise manipulate the hash etc.

jonspalmer:
@anvox See my solution in #518. I believe this is a safe and very general purpose solution for all the use cases. It uses a client option vs subclassing AccessToken which is a major concern with this approach. Also I think its safe for 1-4-stable.

lib/oauth2/access_token.rb
anvox: So, how about support both? :)

anvox:
I like the approach in #518 proc is more flexible than subclass in this case.

PR #518, opened 2020-07-13
Add extract_access_token option to OAuth2::Client

jonspalmer:
Make `OAuth2::Client` responsible for parsing the access_token from the token response by adding an `extract_access_token` option.
This enables clients to supply their own extraction logic so that we can support various use cases like Pintrest and Microsoft that have their own particular wrinkles. 

This is a combination of ideas #516 and #517 but based of the 1-4-stable branch and in a way that is a non-breaking change.

There will be a little bit of work to merge this into master but there is a clear path there too.

### DIFF
--- a/lib/oauth2/access_token.rb
+++ b/lib/oauth2/access_token.rb
@@ -3,6 +3,7 @@ module OAuth2
     attr_reader :client, :token, :expires_in, :expires_at, :params
     attr_accessor :options, :refresh_token
 
+    # Should these methods be deprecated?
     class << self
       # Initializes an AccessToken from a Hash
       #
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -4,7 +4,7 @@ require 'logger'
 module OAuth2
   # The OAuth2::Client class
   class Client # rubocop:disable Metrics/ClassLength
-    RESERVED_PARAM_KEYS = ['headers', 'parse'].freeze
+    RESERVED_PARAM_KEYS = %w[headers parse].freeze
 
     attr_reader :id, :secret, :site
     attr_accessor :options
@@ -25,8 +25,8 @@ module OAuth2
     # @option opts [Symbol] :auth_scheme (:basic_auth) HTTP method to use to authorize request (:basic_auth or :request_body)
     # @option opts [Hash] :connection_opts ({}) Hash of connection options to pass to initialize Faraday with
     # @option opts [FixNum] :max_redirects (5) maximum number of redirects to follow
-    # @option opts [Boolean] :raise_errors (true) whether or not to raise an OAuth2::Error
-    #  on responses with 400+ status codes
+    # @option opts [Boolean] :raise_errors (true) whether or not to raise an OAuth2::Error on responses with 400+ status codes
+    # @option opts [Proc] :extract_access_token  proc that extracts the access token from the response
     # @yield [builder] The Faraday connection builder
     def initialize(client_id, client_secret, options = {}, &block)
       opts = options.dup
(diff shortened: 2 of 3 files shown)


lib/oauth2/access_token.rb
```diff
@@ -3,6 +3,7 @@ class AccessToken
     attr_reader :client, :token, :expires_in, :expires_at, :params
     attr_accessor :options, :refresh_token
 
+    # Should these methods be deprecated?
```
jonspalmer, line 6: I'm not sure what we want to do with these. We could remove them. The `from_kvform` in particular isn't used at all. Seems to be a convenience method. if you're handling token's yourself. I'd be all for dropping them in 2.x  and making this class simpler and general purpose with no assumptions about how the access_token is provided in the hash.

lib/oauth2/client.rb
```diff
@@ -4,7 +4,7 @@
 module OAuth2
   # The OAuth2::Client class
   class Client # rubocop:disable Metrics/ClassLength
-    RESERVED_PARAM_KEYS = ['headers', 'parse'].freeze
+    RESERVED_PARAM_KEYS = %w[headers parse].freeze
```
jonspalmer, line 7: make Rubcop happy. Could make this a separate change if folks would prefer.

lib/oauth2/client.rb
```diff
@@ -93,7 +97,7 @@ def token_url(params = nil)
     #   code response for this request.  Will default to client option
     # @option opts [Symbol] :parse @see Response::initialize
     # @yield [req] The Faraday request
-    def request(verb, url, opts = {}) # rubocop:disable CyclomaticComplexity, MethodLength, Metrics/AbcSize
+    def request(verb, url, opts = {}) # rubocop:disable CyclomaticComplexity, MethodLength, Metrics/AbcSize, Metrics/PerceivedComplexity
```
jonspalmer, line 100: robocop

lib/oauth2/client.rb
```diff
@@ -219,4 +226,26 @@ def redirection_params
       end
     end
   end
+
+  DEFAULT_EXTRACT_ACCESS_TOKEN = proc do |client, hash|
+    token = hash.delete('access_token') || hash.delete(:access_token)
+    token && AccessToken.new(client, token, hash)
+  end
```
jonspalmer, line 233: this is the logic we had in `AccessToken::from_hash`

lib/oauth2/access_token.rb
anvox: I use it to store/restore access token. You could mark it deprecated after providing another better solution, otherwise, leaving a comment like this will confuse other. Thanks.

lib/oauth2/client.rb
```diff
@@ -25,23 +25,27 @@ class Client # rubocop:disable Metrics/ClassLength
     # @option opts [Symbol] :auth_scheme (:basic_auth) HTTP method to use to authorize request (:basic_auth or :request_body)
     # @option opts [Hash] :connection_opts ({}) Hash of connection options to pass to initialize Faraday with
     # @option opts [FixNum] :max_redirects (5) maximum number of redirects to follow
-    # @option opts [Boolean] :raise_errors (true) whether or not to raise an OAuth2::Error
-    #  on responses with 400+ status codes
+    # @option opts [Boolean] :raise_errors (true) whether or not to raise an OAuth2::Error on responses with 400+ status codes
+    # @option opts [Proc] :extract_access_token  proc that extracts the access token from the response
```
anvox, line 29: Please add more comment to remind this proc returning an AccessToken/AccessToken-like object.

lib/oauth2/access_token.rb
jonspalmer: Yup the comment was to draw attention to it for the purpose of this review. I'm happy to just remove this comment if we think these methods stay. It still feels like these methods belong elsewhere.

lib/oauth2/client.rb
jonspalmer: Is there a known pattern for Yard/RDoc to document "Callable" params/options like this? I think we want to declare the signature and the return value. Can just do it in text but I don't know if there is a better way.

anvox:
Thanks. I see it's much better than inheriting AccessToken every time. 
From my view, I still prefer field mapping, I'm testing if we could have it besides proc. 

@pboling what do you think? Could we cut here for 1.4.5, I see many questions/requests on this improvement.

jonspalmer:
As I said previously I'm a bit nervous about the field mapping case because:
 * It will be hard to get the mutation/removal of the access tokens when we find them
 * It's confusing to have two client options that do very similar things

Perhaps the compromise is to have a util method that takes the field mappings and generates the proc for you. So you'd use it like this:

```ruby
extract_access_token = OAuth2::Client.extract_access_token_from_mappings(access_token: ['data', 'access_token'])
client = OAuth2::Client.new('id', 'secret', extract_access_token: extract_access_token)
```

Thoughts? I'd love to cut 1.4.5 with this.

lib/oauth2/access_token.rb
pboling: Agreed on all counts.

lib/oauth2/client.rb
pboling: Rubocop updates are always great.  Clean things up bit by bit as we go.

pboling:
@jonspalmer @anvox I like the idea on having a utility to create the proc that does the mapping.

I will work on cutting 1.4.5, and then we need to push this up to master as well, for 2.0.0 release.

pboling:
@jonspalmer We really need this to be upstreamed to `master`.  Any ideas on how to accomplish that?

PR #532, opened 2021-03-02
Fix initialization of access token expires_at with string value

Lomey:


Lomey:
@pboling, please review & merge if everything is okay. I cannot set up any reviewers or assignees, do not why.

spec/oauth2/access_token_spec.rb
```diff
@@ -70,13 +70,27 @@ def assert_initialized_token(target) # rubocop:disable Metrics/AbcSize
...
+      let(:expires_at) { 1361396829 }
+      let(:hash) do
+        {
+          :access_token => token,
+          :expires_at => expires_at.to_s,
+          'foo' => 'bar'
+        }
+      end
+
+      it 'initializes with a string timestamp expires_at' do
```
dobon, line 83: Add spec for integer expires_at, as well as strings:
```
it 'initializes with a integer time expires_at' do
  target = described_class.from_hash(client, hash.merge(:expires_at => expires_at))
  assert_initialized_token(target)
  expect(target.expires_at).to eql(expires_at)
end
```

lib/oauth2/access_token.rb
```diff
@@ -1,5 +1,7 @@
 module OAuth2
   class AccessToken
+    TIMESTAMP_REGEXP = /^(\d{10,})$/
```
dobon, line 3: unix timestamps don't have to be 10+ digits. They could be less than 10 digits. I think it would be better to rely on the Time library's iso8601 parser instead of this ad hoc timestamp regex.

The code I recommended on the PR discussion also handles integer strings as well as integers in the same case:
```
def convert_expires_at(expires_at)
   begin
      Time.iso8601(expires_at).to_i
   rescue ArgumentError
      expires_at.to_i
   end      
end
```

pboling:
@Lomey Please make the changes recommended by @Lomey .

Meanwhile I am going to do some basic gem housekeeping.

lib/oauth2/access_token.rb
Lomey: did it with regex just because it is much much faster than rescue statement, but okay will change to rescue

Lomey: okay but expires_at could be an integer timestamp(1361396829) or string timestamp("1361396829") and Time.iso8601 method accepts only time strings like this "2011-10-05T22:26:12-04:00", so how does it should work?

dobon: Ahh, good catch. Based on my testing, we'll need to add `TypeError` to the rescue clause to address this:
```
[1] pry(main)> Time.iso8601('2011-10-05T22:26:12-04:00').to_i
=> 1317867972
[2] pry(main)> Time.iso8601('1361396829').to_i
ArgumentError: invalid date: "1361396829"
from ~/.rbenv/versions/2.7.2/lib/ruby/2.7.0/time.rb:639:in `xmlschema'
[3] pry(main)> Time.iso8601(1361396829).to_i
TypeError: no implicit conversion of Integer into String
from ~/.rbenv/versions/2.7.2/lib/ruby/2.7.0/time.rb:616:in `xmlschema'
[4] pry(main)> '1361396829'.to_i
=> 1361396829
[5] pry(main)> 1361396829.to_i
=> 1361396829
```
Important points are that `Time.iso8601` will throw either an `ArgumentError` or `TypeError`, and `to_i` works fine on both strings and integers. So we can do something like this:
```
def convert_expires_at(expires_at)
   begin
      Time.iso8601(expires_at).to_i
   rescue ArgumentError, TypeError
      expires_at.to_i
   end      
end
```
As far as performance between regex and `Time.iso8601`: I have no idea if this is a significant difference. The Time library uses a regex itself, so I would imagine them to be roughly equally fast: https://github.com/ruby/ruby/blob/68f515cf5ed50c74d5f8e19373ea7f6435b62c4e/lib/time.rb#L616

In any case, my personal preference is for stability/robustness over marginal speed. This is a library function though, so maybe we should do tests to confirm whether the error catching will be a problem at scale @pboling ?

dobon: A different, arguably simpler variation is to force the `Time.iso8601` argument to be a string:
```
def convert_expires_at(expires_at)
   begin
      Time.iso8601(expires_at.to_s).to_i
   rescue ArgumentError
      expires_at.to_i
   end      
end
```
NB: added `to_s` to `iso8601` method's parameter, and removed `TypeError` from rescue list.

dobon:
@Lomey @pboling I've created my own PR to make the requested changes: https://github.com/oauth-xx/oauth2/pull/533

If the new PR is accepted, then this one can be closed.

PR #534, opened 2021-03-10
Bump version to 1.4.5 to release #518

mattboldt:
- I noticed `1-4-stable` contains the feature added in #518, but it's not released in v1.4.4 right now. So here's a PR to bump it.
- I also noticed a different fix, from #243, is in `master`, which is supposed to include v2 features? Shouldn't #518 override this?
- Updated README since `bin/setup` and `bin/console` were deleted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,7 +166,7 @@ implementations:
 
 * Ruby 2.2 - Support ends with version 2.x series
 * Ruby 2.3 - Support ends with version 3.x series
-  - [JRuby 9.1][jruby-9.1] (targets MRI v2.3) 
+  - [JRuby 9.1][jruby-9.1] (targets MRI v2.3)
 * Ruby 2.4 - Support ends with version 4.x series
 * Ruby 2.5 - Support ends with version 5.x series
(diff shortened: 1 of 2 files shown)


pboling:
Thanks @mattboldt !

pboling:
@mattboldt the answer to your question is:

> Yes, #518 must be ported to master, and hasn't been thus far.  There was an attempt at a similar solution merged to master, but it was later reverted.

ref: https://github.com/oauth-xx/oauth2/pull/516#issuecomment-662783939

pboling:
@mattboldt 1.4.5 is released

mattboldt:
@pboling Thank you so much! Do y'all plan to merge `1-4-stable -> master` for v2, or will you be cherry picking certain changes? I could look into getting #518 ported to `master`. I think the default proc should support what v1.4.5 does now.

pboling:
There will be no merges from 1-4-stable to master.  Generally fixes will not be back-ported unless someone wants to do the work.  Maintenance effort will go into preparing 2.0 for release.

PR #537, opened 2021-03-18
Fix crash in all calls to OAuth2::Client#get_token

anderscarling:
`def build_access_token` as not found as it was implemented on OAuth2 instead of OAuth2::Client

This error is quite bad as it will happen on each invocation of `OAuth2::Client#get_token` and is suppressed by the fact that the call to `build_access_token` is wrapped in a rescue discarding the exception.

### DIFF
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -231,27 +231,27 @@ module OAuth2
         {}
       end
     end
-  end
 
-  DEFAULT_EXTRACT_ACCESS_TOKEN = proc do |client, hash|
-    token = hash.delete('access_token') || hash.delete(:access_token)
-    token && AccessToken.new(client, token, hash)
-  end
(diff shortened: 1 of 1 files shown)


anderscarling:
Test failure in a rubocop issue unrelated to this change.

anderscarling:
I'm trying to figure out why the this is not an issue while running tests - it's definitely happening for me in actual use.

anderscarling:
This does not break tests due to:

```
# This is dangerous - HERE BE DRAGONS.
# It allows us to refer to classes without the namespace, but at what cost?!?
# TODO: Refactor to use explicit references everywhere
include OAuth2
```

in spec_helper.rb

pboling:
WOW.  I knew that would bite us eventually.  I added the DRAGONS comment, and knew that when I removed the `include Oauth2` a ton of shit broke, and I never got around to figuring it out...

pboling (approved):
LGTM

PR #538, opened 2021-03-19
Remove reliance on globally included OAuth2 in tests (1-4-stable)

anderscarling:
This global inclusion of the OAuth2 recently introduced big issues in version 1.4.5 of OAuth2, fixed in #537.

This should cause big amount of test failures, but everything is green when it's merged with #537.


anderscarling:
Pushed fixes for rubocop issues. All other failures should be related to #537.

PR #539, opened 2021-03-19
Remove reliance on globally included OAuth2 in tests for version 2.0

anderscarling:
This global inclusion of the OAuth2 recently introduced big issues in version 1.4.5 of OAuth2, fixed in #537.

This is basically a version of #538 but against master instead of 1-4-stable.


pboling (approved):
LGTM

PR #550, opened 2021-06-22
Raise an error if Location header not present for redirection

stanhu:
We saw this when trying to use Azure Government Cloud. When
`https://graph.microsoft.com/v1.0/me` was used instead of
`https://graph.microsoft.us/v1.0/me`, Azure returned a 302 response with
no `Location` header. As a result, the OAuth2 client would mysteriously
try to make a GET request to `http:/`, which would fail in Net::HTTP:

```
NoMethodError (undefined method `include?' for nil:NilClass):
/opt/gitlab/embedded/lib/ruby/2.7.0/net/http.rb:1650:in `addr_port'
/opt/gitlab/embedded/lib/ruby/2.7.0/net/http.rb:1585:in `begin_transport'
/opt/gitlab/embedded/lib/ruby/2.7.0/net/http.rb:1518:in `transport_request'
/opt/gitlab/embedded/lib/ruby/2.7.0/net/http.rb:1492:in `request'
```

### DIFF
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -116,7 +116,13 @@ module OAuth2
           verb = :get
           opts.delete(:body)
         end
-        request(verb, response.headers['location'], opts)
+        location = response.headers['location']
+        if location
+          request(verb, location, opts)
+        else
+          error = Error.new(response)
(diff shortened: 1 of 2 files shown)


pboling (approved):
LGTM

PR #551, opened 2021-08-03
skip coverage on non-mri

ahorek:
we should either enable ```simplecov``` gem on non-mri rubies or skip it

fixes failures on travis
```
LoadError:
  cannot load such file -- simplecov
```


pboling (approved):
LGTM

PR #552, opened 2021-08-03
missing require oauth2/version.rb

ahorek:
I think it should be required explicitely

fixes
```
Failure/Error:
  RSpec.describe OAuth2::Version do
....
NameError:
  uninitialized constant OAuth2::Version
```


pboling (approved):
LGTM

PR #560, opened 2022-01-11
Don't set auth params if they are nil

bouk:
Some oauth providers don't like it if you pass an empty client_secret. According to the spec it may be omitted if it's empty: https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1


pboling (approved):
LGTM

PR #567, opened 2022-02-05
Lock RuboCop minor version

bquorning:
In #337 we locked to a minor version of RuboCop, to avoid PRs suddenly failing when a new version of RuboCop was released. Now it looks like we’re back in the same situation, with e.g. (my) #566 failing because of RuboCop.


pboling (approved):
LGTM

PR #576, opened 2022-02-21
🚑️  Stop rescuing parsing errors

### LINKED ISSUE #167: Hiding exceptions is dangerous

pboling:
Signed-off-by: Peter Boling <peter.boling@gmail.com>

Fixed #167

### DIFF
--- a/lib/oauth2/response.rb
+++ b/lib/oauth2/response.rb
@@ -125,9 +125,9 @@ module OAuth2
 end
 
 OAuth2::Response.register_parser(:xml, ['text/xml', 'application/rss+xml', 'application/rdf+xml', 'application/atom+xml', 'application/xml']) do |body|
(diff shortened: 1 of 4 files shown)


xtagon (approved):
LGTM. Breaking change, but worth it IMO as long as it is a major bump to follow semver.

PR #577, opened 2022-03-11
add one more version of parsing the Content-type

kaczor1984:
Implemented third state of Content-type as it was not working with a value like that.


kaczor1984:
It is not necessary - it's enough to change the:
`'content-type':` to `'content-type' =>`

PR #587, opened 2022-06-03
use SHA256 digest

akostadinov:
This patch is for the 1.4 stable branch because it is used in production.

It updates usage of MD5 to SHA256 as a FIPS approved algorythm.

I understand that MD5 usage is fine in general for this use case. The
problem is that on a FIPS configured system, trying to generate MD5 with
openssl fails.

To enable usage on such systems, any other digest algorythm can be used.
SHA256 seems to be the default choice in the community. For example Rails 7.

### DIFF
--- a/lib/oauth2/mac_token.rb
+++ b/lib/oauth2/mac_token.rb
@@ -62,7 +62,7 @@ module OAuth2
     # @param [String] url the HTTP URL path of the request
     def header(verb, url)
       timestamp = Time.now.utc.to_i
-      nonce = Digest::MD5.hexdigest([timestamp, SecureRandom.hex].join(':'))
+      nonce = Digest::SHA256.hexdigest([timestamp, SecureRandom.hex].join(':'))
 
       uri = URI.parse(url)
 


akostadinov:
The failures don't seem to be related to the code change.

pboling:
Thank you @akostadinov

pboling:
@stanhu We will need to make sure this gets upstreamed as well for v2.

pboling:
Oh, I see this is for the MAC Token.  We actually just decided to drop support for MAC Tokens from the v2 release.  Can you elaborate on your use case @akostadinov ?  Is there real world usage of MAC Tokens?  The latest draft, 05, expired many years ago.

pboling:
If there are live use cases, perhaps we should rethink removal in v2?

akostadinov:
Hi, it is probably fine to drop in v2. Just had the requirement to run on a FIPS restricted system in the short term. But I find it good to drop outdated stuff going forward. Thank you for looking into this.

pboling:
@akostadinov this has been released in 1.4.10.  It is anticipated that it will be the final release of the 1.x.x series.

Issue #248, opened 2016-04-14
[MACToken] Client#get_token's access_token_class parameter is broken


yonekawa:
https://github.com/intridea/oauth2/blob/master/lib/oauth2/client.rb#L141

If we try to set `access_token_class` to `MACToken`, `MACToken#signature` fail with `TypeError: no implicit conversion of Hash into String`.

Because, `from_hash` passes (Client, String, Hash) parameters into constructor.
https://github.com/intridea/oauth2/blob/master/lib/oauth2/access_token.rb#L13

However, MACToken's constructor is (Client, String, String, Hash). 3rd parameter is `secret`.
https://github.com/intridea/oauth2/blob/master/lib/oauth2/mac_token.rb#L30

I think we don't need `access_token_class` parameter because we can use `MacToken.from_access_token`.
We should delete it parameter. How about it?

xtagon:
@yonekawa Is this still an open issue? Asking because this was created over a year ago

pboling:
It does appear to still be an issue as the code in question hasn't changed.

yonekawa:
Yes, the code hasn't changed yet.
https://github.com/oauth-xx/oauth2/blob/master/lib/oauth2/access_token.rb#L14

pboling:
There will not likely be action on this until we get a PR from someone motivated to improve the situation.  The team of maintainers has a lot on our plate preparing for the next release, and this issue is only a bug if the code is used in a particular way.  I haven't used it, but the spec passes.  If there is another use case where this doesn't work someone needs to write a spec.

https://github.com/oauth-xx/oauth2/blob/fb502c755ed8801aa5d4575f18b86a9b7ecf9f5e/spec/oauth2/mac_token_spec.rb

pboling:
Added to 2.0.0 release milestone. I'll try to get this working...

stanhu:
I think we should just drop MAC token support: https://github.com/oauth-xx/oauth2/pull/589

pboling:
@yonekawa Are you aware of a live implementation anywhere that is using MAC Token Draft 05 (latest)?  The draft expired many years ago, and it is looking like we'll just drop support for it with v2, so if you know of any use cases please speak up!

PR #588, opened 2022-06-07
Fix Client#get_token with MACToken

stanhu:
`MACToken#signature` would fail with the `#get_token` method, since
the constructor would omit the secret.

To fix this, we implement a `MACToken#from_hash` to properly parse the response.

Closes https://github.com/oauth-xx/oauth2/issues/248

Signed-off-by: Stan Hu <stanhu@gmail.com>

### DIFF
--- a/lib/oauth2/mac_token.rb
+++ b/lib/oauth2/mac_token.rb
@@ -7,6 +7,19 @@ require 'securerandom'
 
 module OAuth2
   class MACToken < AccessToken
+    # Initializes an MACToken from a Hash
+    #
+    # @param client [Client] the OAuth2::Client instance
+    # @param hash [Hash] a hash of MACToken property values
+    # @return [MACToken] the initialized AccessToken
+    def self.from_hash(client, hash)
+      hash = hash.dup
+      new(client,
+          hash.delete('access_token') || hash.delete(:access_token),
+          hash.delete('secret') || hash.delete(:secret),
+          hash)
+    end
+
     # Generates a MACToken from an AccessToken and secret
     #
     # @param [AccessToken] token the OAuth2::Token instance
--- a/spec/oauth2/client_spec.rb
+++ b/spec/oauth2/client_spec.rb
@@ -511,6 +511,33 @@ RSpec.describe OAuth2::Client do
       end
     end
 
+    context 'when a MACToken is requested' do
+      let(:token_params) do
+        {
+          "access_token": 'SlAV32hkKG',
+          "token_type": 'mac',
+          "expires_in": 3600,
+          "refresh_token": '8xLOxBtZp8',
+          "secret": '23sasd#adf@#',
+          "algorithm": 'hmac-sha-1',
+        }
+      end
+
+      it 'returns a MACToken' do
+        client = stubbed_client do |stub|
+          stub.post('/oauth/token') do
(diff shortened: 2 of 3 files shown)


lib/oauth2/client.rb
```diff
@@ -258,7 +258,17 @@ def authenticator
     #
     # @return [AccessToken] the initialized AccessToken
     def build_access_token(response, access_token_opts, access_token_class)
-      access_token_class.from_hash(self, response.parsed.merge(access_token_opts)).tap do |access_token|
+      token_opts = response.parsed.merge(access_token_opts)
+
+      access_token =
+        if access_token_class == OAuth2::MACToken
+          token = AccessToken.from_hash(self, token_opts)
+          OAuth2::MACToken.from_access_token(token, secret)
```
stanhu, line 266: I think `secret` needs to be something else according to https://tools.ietf.org/id/draft-ietf-oauth-v2-http-mac-02.html.

pboling:
@stanhu Please run `bundle exec rubocop --auto-gen-config` and push the result to fix:
```
lib/oauth2/client.rb:156:5: C: Metrics/PerceivedComplexity: Perceived complexity for get_token is too high. [14/13]
    def get_token(params, access_token_opts = {}, access_token_class = AccessToken) ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

pboling (approved):
LGTM

lib/oauth2/mac_token.rb
```diff
@@ -7,6 +7,19 @@
...
+    # Initializes an MACToken from a Hash
+    #
+    # @param client [Client] the OAuth2::Client instance
+    # @param hash [Hash] a hash of MACToken property values
+    # @return [MACToken] the initialized AccessToken
+    def self.from_hash(client, hash)
+      hash = hash.dup
+      new(client,
+          hash.delete('access_token') || hash.delete(:access_token),
+          hash.delete('secret') || hash.delete(:secret),
```
stanhu, line 19: I think we need to handle `mac_key` and `mac_algorithm` according to https://tools.ietf.org/id/draft-ietf-oauth-v2-http-mac-02.html.

stanhu:
@pboling I'm wondering if we should just drop `MACToken` entirely. I haven't found a way to test this without using a toy server.

As far as I can tell, https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-http-mac-05 never made it as a standard. Every OAuth2/OpenID provider I can find (Okta, Google, Azure, Keycloak, etc.) don't seem to support MAC tokens.
It seems like it's obsoleted with OpenID and JWT.

Oracle seems to support a Java client in https://docs.oracle.com/javame/8.0/api/oauth2/api/com/oracle/oauth2/AccessTokenMAC.html.

What do you think?

stanhu:
Dropping this in https://github.com/oauth-xx/oauth2/pull/589.

pboling:
This could be released as a separate gem, if ever needed.

Support for MAC tokens was added in [2014](https://github.com/oauth-xx/oauth2/commit/75cfcf2758199b280995b3eacded7ab29ddb9577), while the draft was still under active development.

PR #589, opened 2022-06-07
Drop support for MAC tokens

### LINKED ISSUE #248: [MACToken] Client#get_token's access_token_class parameter is broken

stanhu:
MAC tokens never worked properly since the secret was never extracted from the authorization
payload. https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-http-mac-05 never made it as a standard, and as far as I can tell, there is virtually no support on the server side for this either.

Closes #248


pboling:
Agree.  If this support is ever needed by anyone, they can use the code removed as a basis for a plugin gem.

pboling:
@stanhu OK, after more investigation, it is clear that the MAC Token draft is dead and not coming back without some mountains being moved.
The original author (@hueniverse) had a repo for the draft [here](https://github.com/hueniverse/draft-hammer-http-mac), and that repo no longer exists.  Also the author's [blog](http://hueniverse.com/2011/12/sled-yahoo-and-moving-on/), where the draft progress was discussed, is no longer online.

PR #593, opened 2022-06-09
Drop IETF MAC Draft 05 CHANGELOG entry

stanhu:
We dropped MAC token support completely since the draft never made it as
a standard, nor is it supported by any major OAuth provider.


pboling (approved):
LGTM

PR #594, opened 2022-06-09
Fix code coverage for version spec

stanhu:
As explained in https://github.com/simplecov-ruby/simplecov/issues/557, since `oauth2.gemspec` calls `require 'oauth2/version.rb`, SimpleCov runs after that file is loaded. We can just exclude it from the code coverage metrics.

In addition, we can drop version specs that fail on a released tag.


pboling (approved):
LGTM

PR #596, opened 2022-06-10
Use a keyword argument for access_token_class in AccessToken#refresh!

stanhu:
This is to be consistent with `OAuth2::Client#get_token`.


pboling (approved):
LGTM

PR #598, opened 2022-06-14
Fix unparsable data not raised as an error

stanhu:
Previously if an unparsable response were returned with a 200
response, an OAuth2::Error would not be raised even if raised_errors
were enabled.

This cleans up and adds tests for this case.


spec/oauth2/client_spec.rb
```diff
@@ -544,6 +548,27 @@ def self.contains_token?(hash)
...
       end
+
+      context 'when status code is 200' do
+        let(:status_code) { 200 }
+
+        context 'when the request body is not nil' do
+          let(:body) { JSON.dump('access_token' => 'the-token') }
+
+          it 'raises an error' do
+            expect { client.get_token({}) }.to raise_error(OAuth2::Error)
```
stanhu, line 559: Hmm, I think I meant to keep this test case as a 500 error? This does not seem right.

pboling (approved):
Great work!

PR #606, opened 2022-06-24
Move access_token_class parameter into Client constructor

### LINKED ISSUE #605: Ruby 2.7 keyword parameter deprecation warning in OAuth2::Client#get_token
Running `bundle exec exec rspec -w`, I see this deprecation warning:

```
OAuth2::Strategy::Password
  #get_token (json)
/Users/stanhu/github/oauth2/lib/oauth2/strategy/password.rb:25: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/stanhu/github/oauth2/lib/oauth2/client.rb:161: warning: The called method `get_token' is defined here
```

I sort of regret mixing positional and keyword arguments now.

stanhu:
Mixing positional and keyword arguments raises deprecation warnings in
Ruby 2.7. It's cleaner to make access_token_class an option at init time.

Closes #605

### DIFF
--- a/lib/oauth2/client.rb
+++ b/lib/oauth2/client.rb
@@ -31,6 +31,7 @@ module OAuth2
     # @option options [Boolean] :raise_errors (true) whether or not to raise an OAuth2::Error on responses with 400+ status codes
     # @option options [Logger] :logger (::Logger.new($stdout)) which logger to use when OAUTH_DEBUG is enabled
     # @option options [Proc] :extract_access_token proc that takes the client and the response Hash and extracts the access token from the response (DEPRECATED)
+    # @option options [Class] :access_token_class [Class] class of access token for easier subclassing OAuth2::AccessToken, @version 2.0+
     # @yield [builder] The Faraday connection builder
     def initialize(client_id, client_secret, options = {}, &block)
       opts = options.dup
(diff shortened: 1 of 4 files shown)


pboling (approved):
I agree.  We intend to be compatible with Ruby >= 2.2, inclusive of 3+, in this release, without deprecation warnings.

lib/oauth2/client.rb
```diff
@@ -156,9 +158,8 @@ def request(verb, url, opts = {})
     # @param params [Hash] a Hash of params for the token endpoint
     # @param access_token_opts [Hash] access token options, to pass to the AccessToken object
     # @param extract_access_token [Proc] proc that extracts the access token from the response (DEPRECATED)
-    # @param access_token_class [Class] class of access token for easier subclassing OAuth2::AccessToken, @version 2.0+
     # @return [AccessToken] the initialized AccessToken
-    def get_token(params, access_token_opts = {}, extract_access_token = options[:extract_access_token], access_token_class: AccessToken)
+    def get_token(params, access_token_opts = {}, extract_access_token = options[:extract_access_token])
```
pboling, line 162: @stanhu Sorry I didn't notice this earlier, but this doesn't appear to be functional.  There wasn't a test that covered it, so I can't be sure.  I refactored it to move the reference to `options` inside the method, and added specs.  Fix will be in 2.0.3 out soon.

stanhu: @pboling Thanks. What was the issue? I thought this was the same signature used in v1.4.9.

PR #607, opened 2022-06-24
Fix raised error class name in CHANGELOG

zavan:
The gem raises ConnectionError, not ConnectionFailed. See: https://github.com/oauth-xx/oauth2/blob/e96dda8e872b517a59d260254180e87a2c63f311/lib/oauth2/client.rb#L117


pboling (approved):
LGTM

PR #620, opened 2022-07-06
Update v2 docs to highlight auth_scheme changes

swanson:
Attempting to upgrade to v2 and ran into trouble. We are using `omniauth` and a provider specific strategy. We needed to change the strategy to explicitly set `auth_scheme` to `request_body` (the pre-v2 default). It was quite difficult to debug why our OAuth flow started breaking so adding this note to the README to help others in the upgrade process.


pboling (approved):
LGTM

PR #626, opened 2022-07-18
2.0.6 uses different token to 2.0.5

rickselby:
I'll start by saying I'm just a user of the gem, I have no idea how the oauth2 process works, we use it with omniauth and cognito for authentication in AWS.

When upgrading to 2.0.6, our logins broke. On investigation, I see the following in the logs:

```
OAuth2::AccessToken.from_hash: `hash` contained more than one 'token' key (["id_token", "access_token"]); using "id_token".
```

Looking at the code changes from #624, the `access_token` was previously being preferred over the `id_token`, so I'm assuming this is the bug.

I've corrected this and added a test to ensure the gem will still function as it did before.

### DIFF
--- a/lib/oauth2/access_token.rb
+++ b/lib/oauth2/access_token.rb
@@ -18,7 +18,7 @@ module OAuth2
       # @return [AccessToken] the initialized AccessToken
       def from_hash(client, hash)
         fresh = hash.dup
-        supported_keys = fresh.keys & TOKEN_KEY_LOOKUP
+        supported_keys = TOKEN_KEY_LOOKUP & fresh.keys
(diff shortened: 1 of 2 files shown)


pboling (approved):
LGTM

vhiairrassary:
We have encountered the same issue with `omniauth-oauth2` and AWS Cognito, and validated this fix:
- 2.0.5 ✅
- 2.0.6 🔴  
- master ✅ 

Thank you 🥳

pboling:
@vhiairrassary @rickselby Thanks for the reports.  The release 2.0.7 should fix this, and be available now.

PR #670, opened 2025-09-08
Add verb-dependent token mode

mrj:
The [Instagram API ](https://developers.facebook.com/docs/instagram-platform/instagram-api-with-instagram-login) uses `:query` token mode for `GET` requests and `:header` mode for `POST` requests, not only for endpoint calls but for [getting](https://developers.facebook.com/docs/instagram-platform/reference/access_token) and [refreshing](https://developers.facebook.com/docs/instagram-platform/reference/refresh_access_token) long-lived tokens. So instead of a fixed mode, an `AccessToken` needs its mode to to be request-verb-dependent. This PR implement this.

Two things I haven't done:

1. Allow the mode to also be dependent on the path and even other options, and
2. Test an invalid mode. I couldn't work out how to stop RSpec using
```
block_is_expected.to raise_error("invalid :mode option of #{mode}")
````
from earlier in the spec file. Not that important given that the normal mode-switching works.

### DIFF
--- a/lib/oauth2/access_token.rb
+++ b/lib/oauth2/access_token.rb
@@ -132,8 +132,9 @@ You may need to set `snaky: false`. See inline documentation for more info.
     # @option opts [FixNum, String] :expires_in (nil) the number of seconds in which the AccessToken will expire
     # @option opts [FixNum, String] :expires_at (nil) the epoch time in seconds in which AccessToken will expire
     # @option opts [FixNum, String] :expires_latency (nil) the number of seconds by which AccessToken validity will be reduced to offset latency, @version 2.0+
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by pboling** [comment]
@mrj Thanks for this.  Reviewing it now, but just noticing that the three failures are probably due to RuboCop style.

First try:
```
bin/rake rubocop_gradual:autocorrect
```

If it fails you can either fix them, or punt them into the future, and run:
```
bin/rake rubocop_gradual:force_update
```

pboling:
@mrj Also, we are working on things at the same time, so please rebase on latest main!

pboling (approved):
LGTM, just needs rebase and style linting.

pboling:
I'll fix the style in post ;) Merging

PR #676, opened 2025-09-13
👷 Add Apache SkyWalking Eyes

pboling:


.licenserc.yaml
```diff
@@ -0,0 +1,3 @@
+dependency:
```
kezhenxu94, line 1: Hi, license-eye needs to know your project’s license, so it can check if the dependencies’ license is compatible with your project’s license, can you add a header license spdx id like the following:


```suggestion
header:
  license:
    spdx-id: MIT

dependency:
```

Also I found a bug in license-eye about the error handling, see https://github.com/apache/skywalking-eyes/pull/206

